PR #22, opened 2024-06-12
Add predicates

fkrause98:
## Changes:
- Add state flags to the vm state.
- Conditionally execute opcodes depending on instructions.
- Add tests for features.


src/lib.rs
gianbelinche, line 1: Why did you add VMState as a return value? Its never used anywhere

gianbelinche (changes requested):
Just that one comment, everything else is fine

src/lib.rs
```diff
@@ -14,14 +14,20 @@ use zkevm_opcode_defs::ISAVersion;
 use zkevm_opcode_defs::LogOpcode;
 use zkevm_opcode_defs::Opcode as Variant;
 
-pub fn run_program(bin_path: &str) -> U256 {
+pub fn run_program(bin_path: &str) -> (U256, VMState) {
```
gianbelinche, line 17: Why did you add VMState as a return value? Its never used anywhere

fkrause98: I though it would be useful to inspect the VM's final state when testing.
Let's say we want to test if an instruction modifies the "eq" flag, I can't come up with another way to test it.

gianbelinche (approved):

juan518munoz (approved):
LGTM

src/lib.rs
juan518munoz: Agree that being able to inspect the `VMState` is useful for testing, it comes in handy for the (yet to be pushed) PR for [arithmetic instructions](https://github.com/matter-labs/zksync-era/blob/main/docs/specs/zk_evm/vm_specification/zkSync_era_virtual_machine_primer.md#arithmetic-instructions).

PR #25, opened 2024-06-13
Flags modifiers for add and sub

### LINKED ISSUE #21: Implement flag modifiers
An instruction can be modified with "!", for example: sub!, will change the vm flags depending on the result of the operation.

fkrause98:
## Changes:
- Handle modifiers for instructions, eg: `add! `should change flags depending on its result.
- Closes #21.


gianbelinche (approved):
LGTM

src/state.rs
```diff
@@ -40,25 +43,35 @@ impl VMState {
...
     }
 
-    pub fn new_with_flag_state(flag_lt: bool, flag_eq: bool, flag_gt: bool) -> Self {
+    pub fn new_with_flag_state(flag_lt_of: bool, flag_eq: bool, flag_gt: bool) -> Self {
         let registers = [U256::zero(); 15];
         let current_frame = CallFrame::new(vec![]);
         Self {
             registers,
-            flag_lt,
+            flag_lt_of,
```
juan518munoz, line 58: Why the rename?

fkrause98: I originally thought this flag was only "less than", but upon re-reading the spec, I've noticed it's actually  "overflow or less than", hence the rename.

juan518munoz (approved):
LGTM

src/state.rs
juan518munoz: Correct, [here](https://github.com/matter-labs/zksync-era/blob/main/docs/specs/zk_evm/vm_specification/zkSync_era_virtual_machine_primer.md#vm-architecture) it states nothing about overflow, yet it is in the [specification document](https://github.com/matter-labs/zksync-era/blob/main/docs/specs/zk_evm/vm_specification/EraVM_formal_specification.pdf).

PR #26, opened 2024-06-13
Mul and div opcodes

### LINKED ISSUE #3: `Mul` and `Div`

juan518munoz:
closes #3 

chore: split program files into associated opcode dir for clarity


Makefile
```diff
@@ -5,21 +5,21 @@ PROGRAMS_DIR=./programs
 ZKSOLC_YUL_FLAGS=--asm --bin --yul --overwrite
 ZKSOLC_ASM_FLAGS=--zkasm --bin --overwrite
 
-YUL_PROGRAMS = $(wildcard $(PROGRAMS_DIR)/*.yul)
-ASM_PROGRAMS = $(wildcard $(PROGRAMS_DIR)/*.zasm)
-ARTIFACTS_YUL = $(patsubst $(PROGRAMS_DIR)/%.yul, $(ARTIFACTS_DIR)/%.artifacts.yul, $(YUL_PROGRAMS))
-ARTIFACTS_ASM = $(patsubst $(PROGRAMS_DIR)/%.zasm, $(ARTIFACTS_DIR)/%.artifacts.zasm, $(ASM_PROGRAMS))
+YUL_PROGRAMS := $(shell find $(PROGRAMS_DIR) -type f -name "*.yul")
```
fkrause98, line 8: Why was this change needed?

src/op_handlers/mul.rs
```diff
@@ -0,0 +1,8 @@
+use crate::address_operands::{address_operands_read, address_operands_store};
+use crate::{opcode::Opcode, state::VMState};
+
+pub fn _mul(vm: &mut VMState, opcode: Opcode) {
+    let (src0, src1) = address_operands_read(vm, &opcode);
+    let res = src0 * src1;
```
fkrause98, line 6: These are 256 bit numbers, keep in mind the spec specifies the mul instruction multiplies the two numbers modulo 2**512. 
The high and low bits are also returned in 2 separate operands, I think address_operands_read only takes into account the first operand destination, correct me If I'm wrong.
Also, we should check if overflow has occurred with mul_overflow, check #25 for an example on how to do this.

Ideally, this PR should also implement the logic for the flag changes, if specified. You can also check #25 for an example.

tests/integration_test.rs
```diff
@@ -140,6 +133,50 @@ fn test_add_does_not_run_if_gt_is_not_set() {
...
+
+    let second_bin_path = make_bin_path_asm("div_remainder");
+    let (remainder_result, _) = run_program(&second_bin_path);
+
+    // 25 / 6 = 4 remainder 1
+    assert_eq!(quotient_result, U256::from_dec_str("4").unwrap());
+    assert_eq!(remainder_result, U256::from_dec_str("1").unwrap());
+}
+
+#[test]
```
fkrause98, line 173: Good job with these tests.

fkrause98 (changes requested):

Makefile
juan518munoz: Not really needed, it's only to avoid eventually having to many similarly named tests on the same directory
image

src/op_handlers/mul.rs
juan518munoz: You are right, I'll change the function, it's corresponding test and check overflow.

Should we also implement the logic for flag changes in this PR, or leave it for a separate one as it's been done with #25?

fkrause98: Yes, bring the changes from #25 now that is merged.

src/op_handlers/div.rs
```diff
@@ -0,0 +1,8 @@
+use crate::address_operands::address_operands_read;
+use crate::{opcode::Opcode, state::VMState};
+
+pub fn _div(vm: &mut VMState, opcode: Opcode) {
+    let (src0, src1) = address_operands_read(vm, &opcode);
+    vm.set_register(opcode.dst0_index, src0 / src1);
+    vm.set_register(opcode.dst1_index, src0 % src1);
```
gianbelinche, line 7: Here you should use address_operands_store for storing the results, right now it only cares about the dst0, it shouldn't be much different to implement it for dst1

Makefile
fkrause98: Makes sense to me, let's just be 100% sure that it does not bring any unwanted problems when running tests and such.

src/op_handlers/div.rs
juan518munoz: Right, the logic of both `mul` and `div` for `address_operand_store` need to be implemented then.

src/address_operands.rs
```diff
@@ -72,25 +72,57 @@ pub fn address_operands_read(vm: &mut VMState, opcode: &Opcode) -> (U256, U256)
...
+            let offset = opcode.imm1;
+            let res = dst1 + U256::from(offset);
+            vm.set_register(opcode.dst1_index, res);
+            res
+        }
+    }
 }
 
-pub fn address_operands_store(vm: &mut VMState, opcode: &Opcode, res: U256) {
+pub fn address_operands_store(vm: &mut VMState, opcode: &Opcode, res: (U256, Option<U256>)) {
```
gianbelinche, line 118: Instead of passing an Option, wouldn't it be best to just have 2 functions, one for dst0 and the other for dst1, since the implementation is different.
I think this adds noise

src/op_handlers/mul.rs
```diff
@@ -1,8 +1,35 @@
...
+        // Gt is set if both of lt_of and eq are cleared.
+        vm.flag_gt |= !vm.flag_lt_of && !vm.flag_eq;
+    }
+
+    address_operands_store(
+        vm,
+        opcode,
+        (
+            low_bits.try_into().unwrap(),
+            Some(high_bits.try_into().unwrap()),
```
gianbelinche, line 32: Maybe add a comment saying that since we used the u256_mask we already know this unwraps won't fail

gianbelinche (approved):
Good job with the fixes!

src/address_operands.rs
fkrause98: Since this is a very special case, I think it's worth adding another implementation for div and mul, here's my proposal:
```rust
pub fn address_operands_store(vm: &mut VMState, opcode: &Opcode, res: U256) { 
    address_operands(vm, opcode, (res, None))
}
pub fn address_operands_div_mul(vm: &mut VMState, opcode: &Opcode, res: (U256, U256){ 
  address_operands(vm, opcode, (res, Some(U256))
}

fn address_operands(vm: &mut VMState, opcode: &Opcode, res: (U256, Option<U256>)) { ... } 
```

src/op_handlers/mul.rs
```diff
@@ -0,0 +1,35 @@
+use u256::{U256, U512};
+
+use crate::address_operands::{address_operands_read, address_operands_store};
+use crate::{opcode::Opcode, state::VMState};
+
+pub fn _mul(vm: &mut VMState, opcode: &Opcode) {
+    let (src0, src1) = address_operands_read(vm, opcode);
+    let src0 = U512::from(src0);
+    let src1 = U512::from(src1);
+    let res = src0 * src1;
```
fkrause98, line 10: I think it would be wiser to use overflowing multiplication here, even if these are U512 integers, if I'm correct, the overflowing multiplication will return a tuple (result, overflow), where overflow will be true if it has ocurred, I'd also use this overflow boolean value to test for overflow.

tests/integration_test.rs
```diff
@@ -245,21 +250,186 @@ fn test_sub_sets_gt_flag_keeps_other_flags_clear() {
...
+    let quotient_result = vm.get_register(3);
+    let remainder_result = vm.get_register(4);
+
+    // program sets registers 3 and 4 at the beginning, and only changes them if the conditional is met
+    assert_eq!(quotient_result, U256::from_dec_str("1").unwrap());
+    assert_eq!(remainder_result, U256::from_dec_str("1").unwrap());
+}
+
+#[test]
+fn test_more_complex_program_with_conditionals() {
```
fkrause98, line 430: Great job thoroughly testing this, must have taken some time 🚀

fkrause98 (changes requested):
Almost there, I think what's more important here is the function addressing, I think my proposal is a nice way to solve this, but feel free to come up with something else if you want. The only thing I'm more worried about it's that the address operands store signature function should be as clean as possible.

src/op_handlers/mul.rs
juan518munoz: The problem with using `U256::overflow_mul()` is that in case of actual overflow we don't get the high bits of the resulting `U512` value. 

Thereby (correct me if wrong) we would even then have to cast to `U512` and then multiply `src0` and `src1` together so we can get the high bits.

fkrause98: Sorry, I meant U512::overflow_mul. Does your comment still apply?

juan518munoz: Considering we are operating with two values that at most are `U256::MAX`, it is impossible that the resulting multiplication between them overflow `U512::MAX`.

```rs
    let u256_max = U256::MAX;
    println!("u256_max: {}", u256_max);
    let u512_max = U512::MAX;
    println!("u512_max: {}", u512_max);
    let overflow = U512::max_value() - (U512::from(u256_max) * U512::from(u256_max));
    println!("overflow: {}", overflow);
 ```
 
 outputs:
 ```bash
 u256_max: 115792089237316195423570985008687907853269984665640564039457584007913129639935
u512_max: 13407807929942597099574024998205846127479365820592393377723561443721764030073546976801874298166903427690031858186486050853753882811946569946433649006084095
overflow: 231584178474632390847141970017375815706539969331281128078915168015826259279870
```

fkrause98: My bad, you're right, I was strictly thinking about U512 values.

fkrause98 (approved):

PR #28, opened 2024-06-13
Add simple gas tracking

fkrause98:


juan518munoz (approved):
LGTM

src/lib.rs
```diff
@@ -45,6 +45,9 @@ pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256,
         let opcode = vm.get_opcode(&opcode_table);
         if vm.predicate_holds(&opcode.predicate) {
             match opcode.variant {
+                // TODO: Properly handle what happens
+                // when the VM runs out of ergs/gas.
+                _ if vm.gas_left() == 0 => break,
```
juan518munoz, line 50: [Maybe `panic` instead of `break`ing.](https://github.com/matter-labs/zksync-era/blob/main/docs/specs/zk_evm/vm_specification/zkSync_era_virtual_machine_primer.md#basic-instructions:~:text=Instructions%20are%20executed%20one%20after%20another%2C%20and%20every%20instruction%20has%20a%20gas%20cost%20measured%20in%20gas.%20A%20program%20that%20runs%20out%20of%20gas%20panics%20and%20none%20of%20its%20side%20effects%20are%20performed.)

gianbelinche, line 50: Why inside the match?

src/lib.rs
```diff
@@ -82,8 +85,9 @@ pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256,
                 Variant::UMA(_) => todo!(),
             }
         }
-
         vm.current_frame.pc += 1;
+        vm.decrease_gas(&opcode);
+        dbg!(opcode.variant.ergs_price());
```
gianbelinche, line 90: Remove the dbg!

src/state.rs
```diff
@@ -69,6 +74,19 @@ impl VMState {
             flag_gt: false,
             flag_eq: false,
             current_frame: CallFrame::new(vec![]),
+            gas_left: Saturating(DEFAULT_GAS_LIMIT),
+        }
+    }
+
+    pub fn new_with_gas(gas_limit: u32) -> Self {
```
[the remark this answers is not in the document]
gianbelinche: +1

tests/integration_test.rs
```diff
@@ -263,3 +263,19 @@ fn test_add_does_not_modify_set_flags() {
     assert!(final_vm_state.flag_lt_of);
     assert!(final_vm_state.flag_eq);
 }
+#[test]
+fn test_runs_out_of_gas_and_stops() {
+    let bin_path = make_bin_path_asm("add_with_costs");
+    let vm = VMState::new_with_gas(5510);
+    let (result, _) = run_program_with_custom_state(&bin_path, vm);
+    assert_eq!(result, U256::from_dec_str("0").unwrap());
+}
```
gianbelinche, line 272: Maybe add a comment explaining what exactly uses all the gas, and what is not being executed

gianbelinche (changes requested):

src/state.rs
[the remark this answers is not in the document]
fkrause98: Yup, you're right.
We've discussed this with @IAvecilla, and we will eventually have a VMBuilder for this.

src/lib.rs
fkrause98: I get what you mean, but panic-ing in Rust is not the same as panic-ing in the VM, the VM just reverts state, Rust kills the thread, so I think breaking is fine for now.

fkrause98: I thought it was the best place to check on how much gas the VM had left before actually executing an instruction, do you think otherwise?
Any suggestions on where to place this check?

src/lib.rs
fkrause98: Fixed.

tests/integration_test.rs
fkrause98: Done.

src/lib.rs
gianbelinche: I would put it just after the loop, since it doesn't actually use the opcode itself, but its not really that important, this works too

gianbelinche (approved):

Issue #12, opened 2024-06-10
`UMA`


jrchatruc:
- `HeapRead`
- `HeapWrite`
- `AuxHeapRead`
- `AuxHeapWrite`
- `FatPointerRead`
- `StaticMemoryRead`
- `StaticMemoryWrite`

PR #30, opened 2024-06-18
Pointer instructions

gianbelinche:
Closes https://github.com/lambdaclass/era_vm/issues/7
This PR adds `ptr.add`, `ptr.sub`, `ptr.pack` and `ptr.shrink`opcodes

PR #35, opened 2024-06-18
Uma opcodes

gianbelinche:
Closes https://github.com/lambdaclass/era_vm/issues/12
This PR add the opcodes related to UMA, `ld, st, ld.1, ld.2, st.1, st.2` as well as `.inc`variants
It builds on top of https://github.com/lambdaclass/era_vm/pull/30 so that one needs to be merged first


src/lib.rs
```diff
@@ -96,7 +102,15 @@ pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256,
...
                     break;
                 }
-                Variant::UMA(_) => todo!(),
+                Variant::UMA(uma_variant) => match uma_variant {
+                    UMAOpcode::HeapRead => _heap_read(&mut vm, &opcode),
+                    UMAOpcode::HeapWrite => _heap_write(&mut vm, &opcode),
+                    UMAOpcode::AuxHeapRead => _aux_heap_read(&mut vm, &opcode),
+                    UMAOpcode::AuxHeapWrite => _aux_heap_write(&mut vm, &opcode),
+                    UMAOpcode::FatPointerRead => _fat_pointer_read(&mut vm, &opcode),
+                    UMAOpcode::StaticMemoryRead => todo!(),
```
jorbush, line 111: In issue #12, include the `StaticMemoryRead` and `StaticMemoryWrite` opcodes. 
I assume you will do this in another PR, but just to confirm.

gianbelinche: In the VM specification there is no mention of StaticMemoryRead or StaticMemoryWrite, so they won't be implemented.
Also, on the vm2, it is marked as an unimplemented instruction.

src/state.rs
```diff
@@ -250,3 +261,49 @@ impl Stack {
...
+            let old_size = self.heap.len() as u32;
+            self.heap.resize(address as usize + 1, 0);
+            return MEMORY_GROWTH_ERGS_PER_BYTE * (address - old_size + 1);
+        }
+        0
+    }
+
+    pub fn store(&mut self, address: u32, value: U256) {
+        for i in 0..32 {
+            self.heap[address as usize + i] = value.byte(31 - i);
```
fkrause98, line 287: Any reason why we aren't using .to_big_endian and simply extending the vector with it? The spec states this is interpreted as big endian. Page 189, if you need further reference.

src/state.rs
```diff
@@ -250,3 +261,49 @@ impl Stack {
...
+        }
+    }
+
+    pub fn read(&mut self, address: u32) -> U256 {
+        let mut result = U256::zero();
+        for i in 0..32 {
```
fkrause98, line 293: Here you can also use to_big_endian, but I think that would be overkill here, and what you're doing is more clear, so I would leave it as is.

fkrause98 (commented):
Looks good, let's discuss the comments I've made before merging.

Great job with the generated tests and the feature.🚀

src/op_handlers/heap_read.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::value::TaggedValue;
+use crate::{opcode::Opcode, state::VMState};
+
+pub fn _heap_read(vm: &mut VMState, opcode: &Opcode) {
+    let (src0, _) = address_operands_read(vm, opcode);
+    if src0.is_pointer {
+        panic!("Invalid operands for heap_read");
+    }
+
+    if src0.value > U256::from(MAX_OFFSET_TO_DEREF_LOW_U32) {
```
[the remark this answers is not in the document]
gianbelinche: I think its better to leave it as is, first because adding an extra function for 4-5 lines seems too much, even if they repeat, and second because this is going to eventually change when we add proper error handling, and then we will implement it in the best way possible

src/state.rs
gianbelinche: I didn't realize that, however, the best thing I can come up with is
```
let mut bytes: [u8;32] = [0;32];
value.to_big_endian(&mut bytes);
for i in 0..32 {
      self.heap[address as usize + i] = bytes[i];
}
```
I can't just extend the vector, since there is an address were it should be stored

src/state.rs
gianbelinche: Perfect, I leave it as is

src/op_handlers/aux_heap_read.rs
```diff
@@ -0,0 +1,31 @@
...
+    if src0.is_pointer {
+        panic!("Invalid operands for heap_read");
+    }
+
+    if src0.value > U256::from(MAX_OFFSET_TO_DEREF_LOW_U32) {
+        panic!("Address too large for heap_read");
+    }
+    let addr = src0.value.low_u32();
+
+    vm.current_frame.aux_heap.expand_memory(addr + 32); // TODO: Handle ergs cost
```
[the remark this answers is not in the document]
gianbelinche: Done!

fkrause98 (approved):
Great job!

src/state.rs
fkrause98: Yup

fkrause98 (approved):

juan518munoz (approved):
LGTM

PR #36, opened 2024-06-19
Add builder for VM State struct

fkrause98:
## Changes
- Add a simple VM builder to streamline VM instantiation, mostly for tests.

### DIFF
--- a/src/state.rs
+++ b/src/state.rs
@@ -24,7 +24,49 @@ pub struct CallFrame {
     // to support in-memory vs on-disk storage, etc.
     pub storage: HashMap<U256, U256>,
 }
+#[derive(Debug, Clone, Default)]
+pub struct VMBuilder {
+    pub registers: Option<[U256; 15]>,
+    pub flag_lt_of: Option<bool>,
+    pub flag_gt: Option<bool>,
+    pub flag_eq: Option<bool>,
+    pub current_frame: Option<CallFrame>,
+}
+impl VMBuilder {
+    pub fn new() -> VMBuilder {
+        Default::default()
+    }
+    pub fn with_registers(mut self, registers: [U256; 15]) -> VMBuilder {
+        self.registers = Some(registers);
+        self
+    }
+    pub fn with_current_frame(mut self, frame: CallFrame) -> VMBuilder {
+        self.current_frame = Some(frame);
+        self
+    }
+    pub fn eq_flag(mut self, eq: bool) -> VMBuilder {
+        self.flag_eq = Some(eq);
+        self
+    }
+    pub fn gt_flag(mut self, gt: bool) -> VMBuilder {
+        self.flag_gt = Some(gt);
+        self
+    }
+    pub fn lt_of_flag(mut self, lt_of: bool) -> VMBuilder {
+        self.flag_lt_of = Some(lt_of);
+        self
+    }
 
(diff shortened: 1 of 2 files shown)


src/state.rs
```diff
@@ -24,7 +24,49 @@ pub struct CallFrame {
     // to support in-memory vs on-disk storage, etc.
     pub storage: HashMap<U256, U256>,
 }
-
+#[derive(Debug, Clone, Default)]
+pub struct VMBuilder {
```
IAvecilla, line 28: Should we name this `VMStateBuilder` to avoid any confussions?

src/state.rs
```diff
@@ -24,7 +24,49 @@ pub struct CallFrame {
...
+
+    pub fn build(self) -> VMState {
+        VMState {
+            registers: self.registers.unwrap_or_else(|| [U256::zero(); 15]),
+            flag_lt_of: self.flag_lt_of.unwrap_or(false),
+            flag_gt: self.flag_gt.unwrap_or(false),
+            flag_eq: self.flag_eq.unwrap_or(false),
+            current_frame: self.current_frame.unwrap_or_else(|| CallFrame::new(vec![])),
+        }
+    }
```
IAvecilla, line 68: Since all these fields are necessary for the `VMState` to work, at least with a default value, it might be better to not declare the fields of the builder as optional. Instead, we could make a manual implementation of the `Default` trait for the builder, declaring all these defaults used here in the build function and then every function can override the default value. Does that make sense?

src/state.rs
fkrause98: Yeah, that's a good idea.

src/state.rs
fkrause98: I don't think this is such a good idea, let's say we implement the default trait. What would the build method look like?
You should somehow check if the programmer has provided some fields, or not, how do you do this without using Option<T>?
Plus, we would have a minor level of indirection here, since the build method would look like something like this:
```rust
VMState {
  self.provided_field,
  ... Default::default()
  }
```
  And you'd have to go and look into the default implementation to see the default fields.
  If this is not what you meant, feel free to correct me 👍 .

gianbelinche (approved):
LGTM

src/state.rs
IAvecilla: The way that I was thinking was something like this:

```rust
impl Default for VMStateBuilder {
    fn default() -> Self {
        VMStateBuilder {
            // Set all the default values you used in the `unwrap_or()` calls in the `build()` method.
        }
    }
}
```

And then for the `build` function you can do something like:

```rust
pub fn build(self) -> VMState {
    VMState {
        registers: self.registers,
        flag_lt_of: self.flag_lt_of,
        // ...
    }
}
```

You wouldn't need the `Option` since every time you call `VMStateBuilder::default()` or `VMStateBuilder::new()`, you already get the default values, and the methods of the builder will override the necessary fields.

Maybe [this](https://rust-unofficial.github.io/patterns/patterns/creational/builder.html) serves as a good example.

fkrause98: Ok, I see what you mean, sounds great, let me implement it and I'll ping you back.

juan518munoz (approved):
LGTM

IAvecilla (approved):

PR #38, opened 2024-06-19
Add call frames

fkrause98:
## Changes:
- Added call frames (execution context) for each call.
- Added simple far call implementation.
- Added test that runs a recursive programs, generates frames, and expects a certain amount of spent gas.
- Adapted coded to use the current context.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -90,7 +90,7 @@ pub fn run(mut vm: VMState) -> (U256, VMState) {
                     LogOpcode::StorageWrite => {
                         let src0 = vm.get_register(opcode.src0_index);
                         let src1 = vm.get_register(opcode.src1_index);
-                        vm.current_context_mut().storage.insert(src0, src1);
+                        vm.current_context_mut().storage.insert(src0.value, src1.value);
                     }
                     LogOpcode::ToL1Message => todo!(),
                     LogOpcode::Event => todo!(),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,12 +1,17 @@
 mod address_operands;
 mod op_handlers;
 mod opcode;
+mod ptr_operator;
 pub mod state;
-mod value;
+pub mod value;
 
 use op_handlers::add::_add;
 use op_handlers::div::_div;
 use op_handlers::mul::_mul;
+use op_handlers::ptr_add::_ptr_add;
+use op_handlers::ptr_pack::_ptr_pack;
+use op_handlers::ptr_shrink::_ptr_shrink;
+use op_handlers::ptr_sub::_ptr_sub;
 use op_handlers::sub::_sub;
 pub use opcode::Opcode;
 use state::CallFrame;
@@ -17,6 +22,7 @@ use zkevm_opcode_defs::definitions::synthesize_opcode_decoding_tables;
 use zkevm_opcode_defs::ISAVersion;
 use zkevm_opcode_defs::LogOpcode;
 use zkevm_opcode_defs::Opcode as Variant;
+use zkevm_opcode_defs::PtrOpcode;
 
 pub fn program_from_file(bin_path: &str) -> Vec<U256> {
     let program = std::fs::read(bin_path).unwrap();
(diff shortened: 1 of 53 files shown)


src/lib.rs
```diff
@@ -90,18 +100,22 @@ pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256,
...
                     LogOpcode::TransientStorageWrite => todo!(),
                 },
-                Variant::FarCall(_) => todo!(),
+                Variant::FarCall(far_call_variant) => far_call(&mut vm, &far_call_variant),
+                // TODO: This is not how return works. Fix when we have calls between contracts
+                // hooked up.
+                // This is only to keep the context for tests
+                Variant::Ret(_) if vm.running_frames.len() > 1 => {
+                    vm.pop_frame();
+                }
```
gianbelinche, line 109: Isn't it better to add this if inside Ret?
```
Variant::Ret(_) => {
    if vm.running_frames.len() > 1 => {
        vm.pop_frame();
    } else {
        break
    }
}
```

src/op_handlers/far_call.rs
```diff
@@ -0,0 +1,14 @@
+use zkevm_opcode_defs::FarCallOpcode;
+
+use crate::state::VMState;
+
+pub fn far_call(vm: &mut VMState, opcode: &FarCallOpcode) {
+    match opcode {
+        FarCallOpcode::Normal => {
+            let program_code = vm.current_context().code_page.clone();
+            let stipend = vm.current_context().gas_left;
+            vm.push_frame(program_code, stipend.0 / 32);
```
gianbelinche, line 10: This is just a placeholder to test the callframes and not the actual farcall implementation right?

gianbelinche (changes requested):
Remember to fix the Lint

src/op_handlers/far_call.rs
fkrause98: Right, I'll be implementing this on another PR.

src/lib.rs
fkrause98: You're right, I'll change it.

tests/integration_test.rs
```diff
@@ -451,26 +452,38 @@ fn test_more_complex_program_with_conditionals() {
...
 }
 #[test]
 // This test should run out of gas before
 // the program can save a number 3 into the storage.
 fn test_runs_out_of_gas_and_stops() {
     let bin_path = make_bin_path_asm("add_with_costs");
-    let vm = VMStateBuilder::new().gas_left(5510).build();
-    let (result, _) = run_program_with_custom_state(&bin_path, vm);
+    let program_code = program_from_file(&bin_path);
+    let frame = CallFrame::new(program_code, 5510);
```
IAvecilla, line 463: The idea of creating a custom `CallFrame` here is to manually set up the gas, right? The idea for the future is to always use `load_program(...)` function and let the VM generate the new context.

fkrause98: Yup, ideally we should always use load_program.

src/lib.rs
```diff
@@ -55,7 +79,9 @@ pub fn run_program(bin_path: &str) -> (U256, VMState) {
 /// Returns the value stored at storage with key 0 and the final vm state.
 pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256, VMState) {
     let program = program_from_file(bin_path);
-    vm.push_frame(program, DEFAULT_INITIAL_GAS);
+    let storage = RocksDB::open(env::temp_dir()).unwrap();
+    let storage = Rc::new(RefCell::new(storage));
+    vm.push_frame(program, DEFAULT_INITIAL_GAS, storage);
```
gianbelinche, line 84: Doesnt this makes `run_program_in_memory` run with rocksdb?

gianbelinche (changes requested):

gianbelinche (approved):
Great Work!

src/lib.rs
fkrause98: fixed

juan518munoz (approved):
LGTM

PR #41, opened 2024-06-21
Implementing `Shift` instructions

jorbush:
#### Overview

This PR introduces the implementation and comprehensive testing for four bitwise shift and rotate operations: `Shl` (Shift Left), `Shr` (Shift Right), `Rol` (Rotate Left), and `Ror` (Rotate Right). 

#### Progress and Testing Details

Each instruction (`Shl`, `Shr`, `Rol`, `Ror`) has the following:

- **Implementation** (`shift.rs`): Defines the operation's behavior, either shifting or rotating bits. 
- **Tests** (`integration_test.rs`):
  - `test_<op>_asm()`: Verifies basic functionality using assembly code.
  - `test_<op>_stack()`: Checks the operation with stack-based addressing.
  - `test_<op>_conditional_eq_set()`: Confirms the operation sets the EQ flag.
  - `test_<op>_set_eq_flag()`: Ensures the EQ flag is set correctly for edge cases where the result is zero.
  -  `test_<op>_asm_greater_than_256()`: Checks what happens when shifting with amounts greater than 256.

#### Conclusion

This PR ensures that the shift and rotate operations (`Shl`, `Shr`, `Rol`, `Ror`) are implemented and tested across various scenarios.

### DIFF
--- a/programs/rol_stack.zasm
+++ b/programs/rol_stack.zasm
@@ -1,0 +1,17 @@
+	.text
+	.file	"rol_stack.zasm"
+	.globl	__entry
+__entry:
+.func_begin0:
+	; grow stack
+	add 1, r0, stack+=[2]
+	; set stack value
+	add 1, r0, stack=[0]
+	add	4, r0, r2
+	; left rotation
+	rol stack=[0], r2, r3
+	sstore r0, r3
+	ret
+.func_end0:
+	.note.GNU-stack
+	.rodata
--- a/programs/rol_stack.zasm
+++ b/programs/rol_stack.zasm
@@ -6,10 +6,10 @@ __entry:
 	; grow stack
 	add 1, r0, stack+=[2]
 	; set stack value
-	add 1, r0, stack=[0]
+	add 1, r0, stack[0]
 	add	4, r0, r2
 	; left rotation
-	rol stack=[0], r2, r3
+	rol stack[0], r2, r3
 	sstore r0, r3
 	ret
 .func_end0:
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -68,7 +68,14 @@ pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256, 
                 Variant::Div(_) => _div(&mut vm, &opcode),
                 Variant::Jump(_) => todo!(),
                 Variant::Context(_) => todo!(),
-                Variant::Shift(_) => _shift(&mut vm, &opcode),
+                Variant::Shift(_) => _shift(
+                    &mut vm,
+                    &opcode,
+                    match opcode.variant {
+                        Variant::Shift(shift_type) => shift_type,
+                        _ => unreachable!(),
+                    },
+                ),
                 Variant::Binop(_) => todo!(),
                 Variant::Ptr(ptr_variant) => match ptr_variant {
                     PtrOpcode::Add => _ptr_add(&mut vm, &opcode),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,6 +7,7 @@ mod value;
 use op_handlers::add::_add;
 use op_handlers::div::_div;
 use op_handlers::mul::_mul;
+use op_handlers::shift::_shift;
 use op_handlers::sub::_sub;
 pub use opcode::Opcode;
 use state::VMState;
@@ -61,7 +62,7 @@ pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256, 
                 Variant::Div(_) => _div(&mut vm, &opcode),
                 Variant::Jump(_) => todo!(),
                 Variant::Context(_) => todo!(),
-                Variant::Shift(_) => todo!(),
+                Variant::Shift(_) => _shift(&mut vm, &opcode),
                 Variant::Binop(_) => todo!(),
                 Variant::Ptr(_) => todo!(),
                 Variant::NearCall(_) => todo!(),
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,7 +12,10 @@ use op_handlers::ptr_add::_ptr_add;
 use op_handlers::ptr_pack::_ptr_pack;
 use op_handlers::ptr_shrink::_ptr_shrink;
 use op_handlers::ptr_sub::_ptr_sub;
-use op_handlers::shift::_shift;
+use op_handlers::shift::_rol;
+use op_handlers::shift::_ror;
+use op_handlers::shift::_shl;
+use op_handlers::shift::_shr;
 use op_handlers::sub::_sub;
 pub use opcode::Opcode;
 use state::VMState;
@@ -22,6 +25,7 @@ use zkevm_opcode_defs::ISAVersion;
 use zkevm_opcode_defs::LogOpcode;
 use zkevm_opcode_defs::Opcode as Variant;
 use zkevm_opcode_defs::PtrOpcode;
+use zkevm_opcode_defs::ShiftOpcode;
 
 /// Run a vm program with a clean VM state.
 pub fn run_program(bin_path: &str) -> (U256, VMState) {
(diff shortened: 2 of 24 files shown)


jorbush:
For ease of review, these are the specifications:

##### `Shl` (Shift Left)
- **Syntax**:
  - `shl in1, in2, out`
  - `shl.s in1, in2, out,` to set swap modifier.
  - `shl! in1, in2, out,` to set set flags modifier.
  - `shl.s! in1, in2, out,` to set both swap and set flags modifiers.
- **Semantic**:
  - Result is computed as `in1 << (in2 % 256)`.
  - Flags are computed as follows:
    - EQ is set if result = 0.
    - Other flags are reset.

##### `Shr` (Shift Right)
- **Syntax**:
  - `shr in1, in2, out`
  - `shr.s in1, in2, out,` to set swap modifier.
  - `shr! in1, in2, out,` to set set flags modifier.
  - `shr.s! in1, in2, out,` to set both swap and set flags modifiers.
- **Semantic**:
  - Result is computed as `in1 >> (in2 % 256)`.
  - Flags are computed as follows:
    - EQ is set if result = 0.
    - Other flags are reset.

##### `Rol` (Rotate Left)
- **Syntax**:
  - `rol in1, in2, out`
  - `rol.s in1, in2, out,` to set swap modifier.
  - `rol! in1, in2, out,` to set set flags modifier.
  - `rol.s! in1, in2, out,` to set both swap and set flags modifiers.
- **Semantic**:
  - Result is computed as `in1 <<< (in2 % 256)`.
  - Flags are computed as follows:
    - EQ is set if result = 0.
    - Other flags are reset.

##### `Ror` (Rotate Right)
- **Syntax**:
  - `ror in1, in2, out`
  - `ror.s in1, in2, out,` to set swap modifier.
  - `ror! in1, in2, out,` to set set flags modifier.
  - `ror.s! in1, in2, out,` to set both swap and set flags modifiers.
- **Semantic**:
  - Result is computed as `in1 >>> (in2 % 256)`.
  - Flags are computed as follows:
    - EQ is set if result = 0.
    - Other flags are reset.

programs/rol_stack.zasm
```diff
@@ -0,0 +1,17 @@
+	.text
+	.file	"rol_stack.zasm"
+	.globl	__entry
+__entry:
+.func_begin0:
+	; grow stack
+	add 1, r0, stack+=[2]
+	; set stack value
+	add 1, r0, stack=[0]
```
gianbelinche, line 9: Right now stack=[] is not working correctly, it is interpreted as stack[] , in order to avoid future errors if this changes, please change all stack=[] for stack[]

tests/integration_test.rs
```diff
@@ -977,3 +977,159 @@ fn test_ptr_pack_in_stack() {
         U256::from_str_radix("0x100000000000000000000000000000000", 16).unwrap()
     );
 }
+
+#[test]
+fn test_shl_asm() {
+    let bin_path = make_bin_path_asm("shl");
+    let r1 = TaggedValue::new_raw_integer(U256::from(1));
+    let r2 = TaggedValue::new_raw_integer(U256::from(2));
```
gianbelinche, line 985: In your shl.zasm you have

```
add 1, r0, r1
add 2, r0, r2
```
So this doesn't actually change anything (If it said for example 3 and 4, the result would not change)
The same goes for the similar tests

gianbelinche (changes requested):

gianbelinche (approved):
LGTM!

src/op_handlers/shift.rs
```diff
@@ -0,0 +1,75 @@
+use crate::address_operands::{address_operands_read, address_operands_store};
+use crate::value::TaggedValue;
+use crate::{opcode::Opcode, state::VMState};
+use zkevm_opcode_defs::Opcode as Variant;
+use zkevm_opcode_defs::ShiftOpcode;
+
+fn _shl(vm: &mut VMState, opcode: &Opcode) {
+    let (src0_t, src1_t) = address_operands_read(vm, opcode);
+    let (src0, src1) = (src0_t.value, src1_t.value);
+    let shift = (src1.low_u32() % 256) as usize;
```
fkrause98, line 10: Not saying this is _wrong_ per se, but I'm a bit confused.

The spec states that you take the lowest byte of the second input, here, you're taking the lowest 4 bytes, are we sure we aren't allowing something that shouldn't work based on the spec? We can also check how the fast vm is doing this to be completely sure. 

Plus, I think the usize cast is redundant, why can't it simply be a u32?

src/op_handlers/shift.rs
```diff
@@ -0,0 +1,75 @@
...
+        vm.flag_gt = false;
+    }
+    address_operands_store(vm, opcode, TaggedValue::new_raw_integer(res));
+}
+
+fn _shr(vm: &mut VMState, opcode: &Opcode) {
+    let (src0_t, src1_t) = address_operands_read(vm, opcode);
+    let (src0, src1) = (src0_t.value, src1_t.value);
+    let shift = (src1.low_u32() % 256) as usize;
+    let res = src0 >> shift;
```
fkrause98, line 26: Same as above

src/op_handlers/shift.rs
```diff
@@ -0,0 +1,75 @@
...
+        vm.flag_gt = false;
+    }
+    address_operands_store(vm, opcode, TaggedValue::new_raw_integer(res));
+}
+
+fn _rol(vm: &mut VMState, opcode: &Opcode) {
+    let (src0_t, src1_t) = address_operands_read(vm, opcode);
+    let (src0, src1) = (src0_t.value, src1_t.value);
+    let shift = (src1.low_u32() % 256) as usize;
+    let result = (src0 << shift) | (src0 >> (256 - shift));
```
fkrause98, line 41: Too bad U256 doesn't implement rotate_left/rotate_right. Looks like this works, nice.

src/op_handlers/shift.rs
```diff
@@ -0,0 +1,75 @@
...
+        vm.flag_eq |= result.is_zero();
+        // other flags are reset
+        vm.flag_lt_of = false;
+        vm.flag_gt = false;
+    }
+    address_operands_store(vm, opcode, TaggedValue::new_raw_integer(result));
+}
+
+pub fn _shift(vm: &mut VMState, opcode: &Opcode) {
+    match opcode.variant {
```
fkrause98, line 68: Make this function simply receive the variant and match it on it directly to avoid the panic!, like this:
```rust
pub fn _shift(vm: &mut, VMState, opcode: &Opcode, type: ShiftOpcode) {
 ....
}
````

Alternatively, you could also change the `panic!` for an [unreachable!](https://doc.rust-lang.org/std/macro.unreachable.html) statement.

File: tests/integration_test.rs [hunk 7]
```diff
@@ -977,3 +977,131 @@ fn test_ptr_pack_in_stack() {
...
+
+#[test]
+fn test_rol_set_eq_flag() {
+    let bin_path = make_bin_path_asm("rol_sets_eq_flag");
+    let (_, vm) = run_program(&bin_path);
+
+    assert!(vm.flag_eq);
+}
+
+#[test]
```
fkrause98, line 1101: Thanks for thoroughly testing this feature 🚀 .

fkrause98 (changes requested):
Overall, it looks good, let's discus a bit more some minor things and we're done.

src/op_handlers/shift.rs
jorbush: Thank you.
At first, I was using `src0.0.rotate_left()` but since it didn't seem to work, I debugged it and it didn't do any rotation (even though I had declared it as mutable).

src/op_handlers/shift.rs
jorbush: Fixed! Thank you 👍

src/op_handlers/shift.rs
jorbush: I checked [fastVM](https://github.com/matter-labs/vm2/blob/9342db726462b76aa7e4ed246684b1316ea79c21/src/instruction_handlers/binop.rs#L106) and if we want something similar, we can use something like this instead:
```rust
let shift = src1.low_u32() as u8;
```
You are right that I get the last 4 bytes, but then I do the `mod 256` and I get the last bytes. I've done that so it's more aligned with the spec: 
```
- The result is computed as in1 << (in2 % 256).
```
But I don't know the performance or readability of the code in this case, should I change it to the above alternative?

src/lib.rs
```diff
@@ -67,7 +68,14 @@ pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256,
...
                 Variant::Context(_) => todo!(),
-                Variant::Shift(_) => todo!(),
+                Variant::Shift(_) => _shift(
+                    &mut vm,
+                    &opcode,
+                    match opcode.variant {
+                        Variant::Shift(shift_type) => shift_type,
+                        _ => unreachable!(),
+                    },
+                ),
```
juan518munoz, line 78: Maybe we should expose `_shl`, `_shr`, `_rol` and `_ror` as public functions and match to them here, so we keep this `match` segment similar to `Variant::Ptr()` and `Variant::Log()`.

jorbush: Done! Thank you!!

src/op_handlers/shift.rs
fkrause98: > But I don't know the performance or readability of the code in this case, should I change it to the above alternative?

As a rule of thumb, let's focus on simplicity+readability more than performance for now 🙂.

`- The result is computed as in1 << (in2 % 256)`
I'm a bit confused by the spec, it states you take the lowest byte as binary digits, so the max amount of shifts you can do
is 255, which is the range for numbers modulo 256. So the in2 % 256 looks a bit redundant. 
But for now, I think following the spec as is is the best thing to do, so let's only take the second argument modulo 256,
and ignore the low 4 bytes, as it seems redundant.
Let's double check this with more people though, pinging @jrchatruc @IAvecilla @gianbelinche.
Let's also have tests checking what happens when shifting with amounts greater than 256.

gianbelinche: I think both `src1.low_u32()` and `src1 % 256` would work.
I don't think the spec is redundant, since it asumes that in2 is a U256, so doing `in2 % 256` is actually the way of getting the last byte

fkrause98 (approved):

PR #44, opened 2024-06-25
Context opcodes

juan518munoz:


src/state.rs
```diff
@@ -33,7 +33,13 @@ pub struct CallFrame {
     pub storage: Rc<RefCell<dyn Storage>>,
     /// Transient storage should be used for temporary storage within a transaction and then discarded.
     pub transient_storage: InMemory,
+    pub context_u128: u128,
+    pub this_address: Address,
+    pub msg_sender: Address,
+    pub code_address: Address,
+    pub ergs_remaining: u32,
```
gianbelinche, line 40: Maybe add all the new values in a new struct in order to make it cleaner?

juan518munoz: [This fields are now part of the `CallFrame` struct](https://github.com/lambdaclass/era_vm/blob/2562a1f2d100f1fbb3d7debe5a9c487ff5930858/src/call_frame.rs#L29-L36), I cant try moving them into a 
new `Context` struct so that `CallFrame` doesn't get saturated.

gianbelinche: Try to name it something else than `Context` since Near Call already has a `Context` Struct, so we don't overlap later

gianbelinche: We can choose later which of the 2 should actually be called `Context`

juan518munoz: Looking at the [`Context` struct](https://github.com/lambdaclass/era_vm/pull/48/files#diff-f45de6772e2dbec81333a31f1f67583428ef2353e386df23dbbe69ba83a48c37R29-R33) being introduced in #48, I'm wondering if it wouldn't be more suitable to add these new fields that struct considering that it's share beetwen the `frame` and `near_call_frames`.

Am I'm missing something?

gianbelinche: It depends, do these new field change when calling a contract with near or far call? If they don't, then its fine to put them on `Context`, if they do, they belong on `CallFrame`

src/.DS_Store
jrchatruc, line 1: Remove this file (maybe also add `.DS_Store` to `.gitignore`)

juan518munoz: Done!

.gitignore
```diff
@@ -2,3 +2,4 @@
 program_artifacts
 /scripts/*.zasm
 /tests/test_storage
+src/.DS_Store
```
Oppen, line 5:
```suggestion
**/.DS_Store
```

programs/context_code_address.zasm
Oppen, line 1: Would make sense for the name to match the instruction, no?

programs/context_code_ergs_left.zasm
Oppen, line 1: Same here.

Oppen:
There seems to be mixed indentation (4 spaces and tabs) in the `.zasm` files.

src/call_frame.rs
```diff
@@ -32,16 +37,26 @@ pub struct Context {
 }
 
 impl Context {
-    pub fn new(program_code: Vec<U256>, gas_stipend: u32) -> Self {
+    pub fn new(
+        program_code: Vec<U256>,
+        gas_stipend: u32,
+        address: Address,
```
Oppen, line 43: Whose address is this? The contract's?

src/op_handlers/context.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub fn _caller(vm: &mut VMState, opcode: &Opcode) {
+    let res = TaggedValue::new_raw_integer(address_to_u256(&vm.current_frame().caller));
+    address_operands_store(vm, opcode, res);
+}
+
+pub fn _code_address(vm: &mut VMState, opcode: &Opcode) {
+    let res = TaggedValue::new_raw_integer(address_to_u256(&vm.current_frame().code_address));
+    address_operands_store(vm, opcode, res);
+}
```
Oppen, line 31: I infer, from the names, that these should be creating pointers rather than integers.

src/op_handlers/context.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub fn _ergs_left(vm: &mut VMState, opcode: &Opcode) {
+    let res = TaggedValue::new_raw_integer(U256::from(vm.current_frame().gas_left.0));
+    address_operands_store(vm, opcode, res);
+}
+
+pub fn _sp(vm: &mut VMState, opcode: &Opcode) {
+    let sp = vm.current_frame().stack.sp();
+    address_operands_store(vm, opcode, TaggedValue::new_raw_integer(U256::from(sp)));
```
Oppen, line 55: The same applies here. The stack pointer is, intuitively, a pointer.

Oppen:
Do the tests test actual behavior or are they just smoke tests?

src/op_handlers/context.rs
gianbelinche: I don't think they are, since `TaggedValue::new_pointer` actually indicates that a Fat Pointer is being stored, which is not the case, if it leads to confusion we can change it to `TaggedValue::new_fat_pointer`

src/call_frame.rs
```diff
@@ -22,6 +23,10 @@ pub struct CallFrame {
     pub gas_left: Saturating<u32>,
     /// Transient storage should be used for temporary storage within a transaction and then discarded.
     pub transient_storage: InMemory,
+    pub address: Address,
+    pub caller: Address,
+    pub code_address: Address,
+    pub context_u128: u128,
```
gianbelinche, line 29: I think this would be better stored on the Context struct, since they do not change between near calls

src/op_handlers/context.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub fn _set_context_u128(vm: &mut VMState, opcode: &Opcode) {
+    let (src0, _) = address_operands_read(vm, opcode);
+    vm.current_frame_mut().context_u128 = src0.value.as_u128();
+}
+
+pub fn _aux_mutating0(_vm: &mut VMState, _opcode: &Opcode) {
+    // unknown behaviour, should not be called
+    panic!("aux_mutating0 called");
+}
```
gianbelinche, line 71: Do you know why this exists?

src/state.rs
```diff
@@ -22,6 +23,16 @@ pub struct Heap {
...
+impl Heap {
+    pub fn len(&self) -> usize {
+        self.heap.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.heap.is_empty()
+    }
+}
+
```
gianbelinche, line 35: There is already an impl Heap down this file, combine both

tests/integration_test.rs
```diff
@@ -678,6 +679,106 @@ fn test_vm_generates_frames_and_spends_gas() {
...
+#[test]
+fn test_context_meta() {
+    let bin_path = make_bin_path_asm("context_meta");
+    let program = program_from_file(&bin_path);
+    let context = Context::new(program, 9999, Address::default(), Address::default());
+    let vm = VMStateBuilder::new().with_contexts(vec![context]).build();
+    let (res, _) = run_program(&bin_path, vm, &mut []);
+    assert_eq!(res, U256::from_dec_str("0").unwrap());
+}
+
```
gianbelinche, line 771: Maybe add a test were meta != 0, for example storing something on heap

gianbelinche (changes requested):

tests/integration_test.rs
juan518munoz: Done!

src/state.rs
juan518munoz: Missed it!

src/op_handlers/context.rs
juan518munoz: No, I couldn't find anything about it on the specification other that [it's not implemented](https://matter-labs.github.io/eravm-spec/spec.html#IntValue:~:text=(*%C2%A0TODO%3A%C2%A0the%C2%A0following%C2%A0is%C2%A0not%C2%A0implemented%C2%A0yet%C2%A0*)%0A%C2%A0%C2%A0%C2%A0%C2%A0%7C%20Assembly.OpAuxMutating%20in1%20%E2%87%92%20%40OpAuxMutating%20decoded%20in1).

src/call_frame.rs
juan518munoz: You are right, I've moved them there.

src/op_handlers/context.rs
juan518munoz: Correct.

src/op_handlers/context.rs
juan518munoz: Yes, this is due to a mistake, it should be named `context_ergs_left.zasm` instead of `context_code_ergs_left.zasm`

programs/context_code_address.zasm
juan518munoz: Tests are named after their respective op codes, the problem with `CodeAddress` is that it's associated instruction differs in naming.
image
[Source](https://matter-labs.github.io/eravm-spec/spec.html)

programs/context_code_ergs_left.zasm
juan518munoz: Yes, this is due to a mistake, it should be named `context_ergs_left.zasm` instead of `context_code_ergs_left.zasm`

src/call_frame.rs
juan518munoz: Correct.

src/op_handlers/context.rs
gianbelinche: Still, the sp should not be stored as a `TaggedValue::new_pointer`, since it is not a Fat Pointer

programs/context_meta.zasm
```diff
@@ -3,6 +3,12 @@
 	.globl	__entry
 __entry:
 .func_begin0:
+	add 42, r0, r1
+	; store 42 in heap incrementing by 33 bytes
+	st.1 r0, r1
+	; store in aux heap incrementing by 33 bytes
+	st.2 r0, r1
```
Oppen, line 10: For that increment you'd be missing the `.inc` modifier I believe. Also, isn't it 32 bytes?

src/op_handlers/context.rs
Oppen: I think that would be sensible.

src/state.rs
juan518munoz: Marking as resolved ([actually continued here](https://github.com/lambdaclass/era_vm/pull/44#discussion_r1669043582))

programs/context_meta.zasm
juan518munoz: >For that increment you'd be missing the .inc modifier I believe.

It goes without any modifier. If your doubt comes from the fact that the spec document lists [`OpStore`](https://matter-labs.github.io/eravm-spec/spec.html#:~:text=in2%20aliased%20as-,st.1.inc%20in1%2C%20out,-uma.aux_heap_write%20in1) using the `inc` modifier, I believe it's an error, as it would make it identical to [`OpStoreInc`](https://matter-labs.github.io/eravm-spec/spec.html#:~:text=in2%20aliased%20as-,st.1.inc%20in1%2C%20out,-uma.inc.aux_heap_write).

>Also, isn't it 32 bytes?

Correct! it appears as there was a small error on our codebase which I interpreted wrongly, it's now fixed.

juan518munoz:
>Do the tests test actual behavior or are they just smoke tests?

Most of them are dumb because we lack the implementation of far calls at the moment, which is what generates the most of `Context` changes. Once that gets implemented we should further develop integration testing for this opcodes.

src/call_frame.rs
Oppen: I'd suggest adding that information to the name: `contract_address`.

src/call_frame.rs
```diff
@@ -29,13 +30,26 @@ pub struct CallFrame {
 pub struct Context {
     pub frame: CallFrame,
     pub near_call_frames: Vec<CallFrame>,
+    pub contract_address: Address,
```
fkrause98, line 33: Please add a a doc comment for each new field here.

src/op_handlers/context.rs
```diff
@@ -0,0 +1,75 @@
...
+
+pub fn set_context_u128(vm: &mut VMState, opcode: &Opcode) {
+    let (src0, _) = address_operands_read(vm, opcode);
+    vm.current_context_mut().context_u128 = src0.value.as_u128();
+}
+
+pub fn aux_mutating0(_vm: &mut VMState, _opcode: &Opcode) {
+    // unknown behaviour, should not be called
+    panic!("aux_mutating0 called");
```
fkrause98, line 70: Change this panic message to 'Undefined behaviour' or 'Not in spec'

src/op_handlers/far_call.rs
```diff
@@ -1,14 +1,21 @@
...
+use crate::{state::VMState, Opcode};
+
+fn u256_to_address(integer: &U256) -> Address {
+    let mut buffer = [0u8; 32];
+    integer.to_big_endian(&mut buffer);
+
+    Address::from_slice(&buffer[12..32])
+}
+
+pub fn _far_call_normal(vm: &mut VMState, opcode: &Opcode) {
```
fkrause98, line 13: Restore this to the far_call function and match on the opcode to call this one.
lib.rs should still call far_call.

tests/integration_test.rs
```diff
@@ -678,6 +680,116 @@ fn test_vm_generates_frames_and_spends_gas() {
...
+        this_shard_id: 0,
+        caller_shard_id: 0,
+        code_shard_id: 0,
+        aux_field_0: 0,
+    })
+    .to_u256();
+    assert_eq!(res, expected);
+}
+
+#[test]
```
fkrause98, line 783: Great job testing this 🚀

fkrause98 (changes requested):

fkrause98:
> Most of them are dumb because we lack the implementation of far calls at the moment, which is what generates the most of Context changes. Once that gets implemented we should further develop integration testing for this opcodes.


@juan518munoz let's make sure we have an issue for that.

src/call_frame.rs
juan518munoz: Done, fyi, `context_u128` doc comes from [here](https://docs.zksync.io/zk-stack/components/zksync-evm/vm-specification#context-value).

src/op_handlers/context.rs
juan518munoz: Let's go with 'Undefined behaviour'.

Oppen: I'm not entirely sure I understand what's going on here. We added this but it's not in spec. Where does it come from?
Whatever this is, maybe a more detailed comment is due, say, if we found some assembly uses it but we don't know for what, state just that. Few-weeks-in-the-future us will be thankful.
Also, I'm not sure "undefined behavior" is appropriate here. It means a much more complicated thing where that term comes from than just something being unspecified. I'd go with "undocumented" instead.

juan518munoz: [This](https://matter-labs.github.io/eravm-spec/spec.html#:~:text=%C2%A0%C2%A0(*%C2%A0TODO%3A%C2%A0the%C2%A0following%C2%A0is%C2%A0not%C2%A0implemented%C2%A0yet%C2%A0*)%0A%C2%A0%C2%A0%C2%A0%C2%A0%7C%20Assembly.OpAuxMutating%20in1%20%E2%87%92%20%40OpAuxMutating%20decoded%20in1) is the only documentation we have on `AuxMutating0`, we might follow as the spec document implies and panic with the message "Not implemented yet"

```rs
pub fn aux_mutating0(_vm: &mut VMState, _opcode: &Opcode) {
    // None of the other vms have the implementation for this opcode
    // from spec document:
    // (* TODO: the following is not implemented yet *)
    panic!("Not yet implemented");
}
```

fkrause98: Let's go with that one then.

src/op_handlers/far_call.rs
fkrause98: No, what I mean is to restore the far_call function like it was previously.
Then, call far_call_normal inside that one.
AFAIK, each far_call variant has the same setup, give or take, so I see value in having everything in one place, at least for now.

juan518munoz: I see, I've reverted this changes, but there are some changes needed tho, like passing the complete `Opcode` as an argument instead of only the type of `FarCallOpcode`, to access fields such as `src0_index`.

```rs
pub fn far_call(vm: &mut VMState, opcode: &Opcode, far_call_variant: FarCallOpcode) {
    match far_call_variant {
        FarCallOpcode::Normal => {
            let program_code = vm.current_frame().code_page.clone();
            let stipend = vm.current_frame().gas_left;
            let address_mask: U256 = U256::MAX >> (256 - 160);
            let address =
                u256_to_address(&((vm.get_register(opcode.src0_index).value) & address_mask));
            let caller = vm.current_context().contract_address;
            vm.push_far_call_frame(program_code, stipend.0 / 32, address, caller);
        }
        _ => todo!(),
    }
}
```

fkrause98 (approved):

src/op_handlers/context.rs
Oppen: I only see `AuxMutating`, which receives an input register, supposedly required to be a pointer. Maybe `aux_mutating0` is meant to handle the case where the input is `r0`, which always returns `0` (which probably makes no sense as a pointer). I'd go with the not implemented message for now, but it's worth asking upstream.

gianbelinche (approved):
LGTM!

Issue #47, opened 2024-06-26
Add RollBacks


gianbelinche:
When doing a near call, the storage needs to be cloned, so that the state is the same, but the changes are not present on the previous callframe.
However, rocksdb cannot be cloned, so in order to implement this behaviour, a rollback mechanism needs to be implemented.
In this new format, the storage would be passed as is, and when the context is restored, the rollback should happend, restoring the previous state.

juan518munoz:
Should we close this issue given that #108 is already in `main` and this [issue](https://github.com/lambdaclass/era_vm/issues/196) aims to get a complete implementation?

gianbelinche:
> Should we close this issue given that #108 is already in `main` and this [issue](https://github.com/lambdaclass/era_vm/issues/196) aims to get a complete implementation?

That is fine, but lets remember that we need to optimize this later, it still going to be a rough implementation of rollbacks (we need to implement something like the rollbackable map the have on vm2)

juan518munoz:
Agree, after the mentioned issue is resolved we should check if a full rollback is implemented, otherwise we should open a new issue.

PR #48, opened 2024-06-26
Near Call Opcode

gianbelinche:
Closes https://github.com/lambdaclass/era_vm/issues/8
This PR adds Near Call opcode.

Now the VM has the following structure related to callframes:

There is a vector of Contexts, each context has a Callframe and a vector of Callframes.
The idea behind this is, that every time a far_call is called, a new Context is pushed to the contexts vector, but if a near call is called, the callframe is pushed to the vector of callframes of the current context.
This is done this way because when a callframe created with a near call is popped, some state needs to be mantained, and others need to be rollbacked.

This PR depends on https://github.com/lambdaclass/era_vm/pull/38

Note to the future:
When creating a new callframe for near call, the storage is being "fake cloned", since it is not possible to clone rocksdb, it is returning an InMemory storage, a new issue was added in order to implement proper rollbacks. https://github.com/lambdaclass/era_vm/issues/47


programs/near_call.zasm
```diff
@@ -0,0 +1,25 @@
...
+.func_begin1:
+	; gas add = 6
+	add 5,r0,r1
+	; gas ret = 5
+	ret
+.func_end1:
+
+
+	.note.GNU-stack
+	.rodata
```
juan518munoz, line 25: Is gas consumption being tracked properly?
```rs
#[test]
fn test_near_call() {
    let bin_path = make_bin_path_asm("near_call");
    let (result, vm) = run_program(&bin_path);
    let expected_gas_left = DEFAULT_INITIAL_GAS - (25 + 5111 + 5 + 6 + 5);
    assert_eq!(expected_gas_left, vm.gas_left());
    assert_eq!(result, U256::from(5));
}
```
Fails:
```bash
assertion `left == right` failed
  left: 60384
 right: 59984
 ```

**Review by juan518munoz — CHANGES_REQUESTED** [review]

File: tests/integration_test.rs [hunk 2]
```diff
@@ -1631,52 +1630,117 @@ fn test_ptr_pack_in_stack() {
     );
 }
 
+#[test]
+fn test_near_call() {
+    let bin_path = make_bin_path_asm("near_call");
+    let (result, _) = run_program(&bin_path);
+    assert_eq!(result, U256::from(5));
```
fkrause98, line 1637: Let's add gas tracking for these tests.

fkrause98 (changes requested):

programs/near_call.zasm
gianbelinche: The `sstore` opcode has a gas cost of 5511, you put 5111, the output value is actually correct, I'll add this test so we make sure it doesn't break

tests/integration_test.rs
gianbelinche: Done!

programs/near_call.zasm
juan518munoz: You are right, did not notice the typo, sorry for the mistake!

juan518munoz (approved):

src/call_frame.rs
```diff
@@ -19,28 +19,62 @@ pub struct CallFrame {
     pub aux_heap: Heap,
     pub code_page: Vec<U256>,
     pub pc: u64,
-    /// Storage for the frame using a type that implements the Storage trait.
-    /// The supported types are InMemory and RocksDB storage.
+    // TODO: Storage is more complicated than this. We probably want to abstract it into a trait
+    // to support in-memory vs on-disk storage, etc.
```
IAvecilla, line 23: I think this is a wrong change after merge, right?

src/call_frame.rs
```diff
@@ -19,28 +19,62 @@ pub struct CallFrame {
...
     pub storage: Rc<RefCell<dyn Storage>>,
-    /// Transient storage should be used for temporary storage within a transaction and then discarded.
-    pub transient_storage: InMemory,
     pub gas_left: Saturating<u32>,
+    pub transient_storage: InMemory,
```
IAvecilla, line 26: Same here, there are some comments missing

src/lib.rs
```diff
@@ -174,12 +175,13 @@ pub fn run(mut vm: VMState) -> (U256, VMState) {
...
-    let final_storage_value = match vm.current_context().storage.borrow().read(&U256::zero()) {
-        Ok(value) => value,
-        Err(_) => U256::zero(),
-    };
+    let final_storage_value = vm
+        .current_frame()
+        .storage
+        .borrow()
+        .read(&U256::zero())
+        .unwrap();
```
IAvecilla, line 185: Is there a reason why we changed the match clause? it doesn't matter that much but just out of curiosity.

src/call_frame.rs
```diff
@@ -19,28 +19,62 @@ pub struct CallFrame {
...
-        storage: Rc<RefCell<dyn Storage>>,
-    ) -> Self {
+    pub fn new_far_call_frame(program_code: Vec<U256>, gas_stipend: u32) -> Self {
         Self {
             stack: Stack::new(),
             heap: Heap::default(),
             aux_heap: Heap::default(),
             code_page: program_code,
             pc: 0,
+            storage: Rc::new(RefCell::new(InMemory::default())),
```
IAvecilla, line 52: Are you doing this because you will override the storage value for all the frames in the context, right? This will probably change a lot with the implementation of the `far_call` opcode but you can add a comment explaining that this value is a default.

src/call_frame.rs
gianbelinche: Yes, probably some copy paste there because of the change of directory or something

src/call_frame.rs
gianbelinche: I'll add it back!

src/lib.rs
gianbelinche: The change of `current_context` for `current_frame` was intentional, since now the `current_context` is actually just the one of the far calls, and current frame could be from a near call.
About the match for the unwrap, it was not...
However I think it's better this way, since a lot of times, some tests failed because they didn't reach the `sstore` instruction, and having the unwrap was a really quick way of knowing why it failed, if it just returned 0, then It would have been more difficult to notice.
So in the end, it was not on purpose, but useful anyways

src/call_frame.rs
gianbelinche: Since I was not implementing far calls, this was just a way of keeping what was already implemented functioning, and yes, if you use later a builder with storage, it will override it.
I'll add the comment

IAvecilla (approved):

fkrause98 (approved):

fkrause98:
@gianbelinche Great work 🚀

PR #50, opened 2024-06-26
Swap modifier

### LINKED ISSUE #46: Add swap modifier (.s) to all instructions that need it
In particular, all ptr instructions ,` ptr.add , ptr.sub, ptr.shrink and ptr.pack` should use it, but when looking at the opcode table, there is not any entrance for this opcodes where the swap_flag is true, this is noted on the [Notion questions](https://www.notion.so/lambdaclass/Era-VM-Docs-and-Questions-6d6c00a49e1744b8a6f8f171f07a4a7e).

juan518munoz:
Closes #46
Implement `.s` modifier.

### DIFF
--- a/src/address_operands.rs
+++ b/src/address_operands.rs
@@ -26,7 +26,7 @@ fn reg_and_imm_read(vm: &mut VMState, opcode: &Opcode) -> (TaggedValue, TaggedVa
 }
 
 pub fn address_operands_read(vm: &mut VMState, opcode: &Opcode) -> (TaggedValue, TaggedValue) {
-    match opcode.src0_operand_type {
+    let (op1, op2) = match opcode.src0_operand_type {
         Operand::RegOnly => only_reg_read(vm, opcode),
         Operand::RegOrImm(variant) => match variant {
             RegOrImmFlags::UseRegOnly => only_reg_read(vm, opcode),
@@ -74,6 +74,11 @@ pub fn address_operands_read(vm: &mut VMState, opcode: &Opcode) -> (TaggedValue,
                 }
             }
         }
+    };
+    if opcode.swap_flag {
+        (op2, op1)
+    } else {
+        (op1, op2)
     }
 }
 
--- a/src/op_handlers/div.rs
+++ b/src/op_handlers/div.rs
@@ -3,7 +3,9 @@ use crate::value::TaggedValue;
 use crate::{opcode::Opcode, state::VMState};
 
 pub fn _div(vm: &mut VMState, opcode: &Opcode) {
+    println!("swap flag: {}", opcode.swap_flag);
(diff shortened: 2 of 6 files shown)


src/op_handlers/div.rs
```diff
@@ -3,7 +3,9 @@ use crate::value::TaggedValue;
 use crate::{opcode::Opcode, state::VMState};
 
 pub fn _div(vm: &mut VMState, opcode: &Opcode) {
+    println!("swap flag: {}", opcode.swap_flag);
```
fkrause98, line 6: Remove these prints.

src/address_operands.rs
```diff
@@ -74,6 +74,11 @@ pub fn address_operands_read(vm: &mut VMState, opcode: &Opcode) -> (TaggedValue,
                 }
             }
         }
+    };
+    if opcode.swap_flag {
```
fkrause98, line 78: Could there be a combination of inputs that when swapped becomes non-valid? Eg: If I do `add.swap 1, r0, r1`

fkrause98 (changes requested):

src/address_operands.rs
```diff
@@ -74,6 +74,11 @@ pub fn address_operands_read(vm: &mut VMState, opcode: &Opcode) -> (TaggedValue,
...
+        (op2, op1)
+    } else {
+        (op1, op2)
```
gianbelinche, line 81: I'm not sure if this is how we should do it, I think not all opcodes support swapping, maybe this should be implemented in each opcode. (Please check this, maybe its fine)

gianbelinche (changes requested):

src/address_operands.rs
juan518munoz: Compiler forbids this kind of instructions:
```asm
	div.s r1, 3, r3, r4
```
throws:
```bash
error: assembly parsing: can not parse text section element: cannot parse lines: {6: ("div.s r1, 3, r3, r4", InvalidOperandImmInRegLocation("3"))}
```

src/address_operands.rs
juan518munoz: As you say, some functions, like `Mul`, don't support the `.s`, this is taken care of by the compiler, which forbids bulding this kind of code:
```asm
	mul.s 3, r1, r3, r4
```
throws:
```bash
 error: assembly parsing: can not parse text section element: cannot parse lines: {6: ("mul.s 3, r1, r3, r4", UnknownArgument("Mul instruction contains unknown modifiers: {\"s\"}"))}
 ```

gianbelinche (approved):

gianbelinche:
LGTM!

fkrause98 (approved):
Looks good. Feel free to merge after solving the merge conflict.

PR #52, opened 2024-06-26
Add far call

fkrause98:
## Changes:
- First iteration of the far call opcode, which calls another contract.
- Structural changes to have decommit and contract storages in the near future.
- Make the storage be a single, global field accessed through the VM Struct.


src/op_handlers/far_call.rs
```diff
@@ -1,13 +1,50 @@
...
+// 1 - Decode the parameters. (done)
+// 2 - Decommit the address. (WIP)
+// 3 - Load the new context. (WIP)
+pub fn far_call(vm: &mut VMState, opcode: &Opcode, far_call: &FarCallOpcode) {
+    let (src0, src1) = address_operands_read(vm, opcode);
+    let contract_address = address_from_u256(&src1.value);
+    let _err_routine = opcode.imm0;
+    let FarCallParams { ergs_passed, .. } = far_call_params_from_register(src0.value);
+    match far_call {
+        FarCallOpcode::Normal if ergs_passed < vm.gas_left() => {
```
gianbelinche, line 44: I think that if `ergs_passed > vm.gas_left()`, you just pass` vm.gas_left()` (or actually 63/64 of it)

tests/integration_test.rs
```diff
@@ -31,7 +34,22 @@ fn make_bin_path_asm(file_name: &str) -> String {
...
     )
 }
-
+struct TestDB {
+    pub ptr: Rc<RocksDB>,
+    db_path: PathBuf,
+}
+impl TestDB {
+    fn new() -> Self {
+        let db_path = PathBuf::from(format!("./.test_db.{}/", fake_rand()));
```
gianbelinche, line 43: Can we make sure this won't collide when running tests?

gianbelinche (changes requested):
I understand this is just a partial implementation of far call, but is it possible to add some tests with what we already have?

tests/integration_test.rs
fkrause98: This struct removes the db file after going out of scope: https://github.com/lambdaclass/era_vm/pull/52/files#diff-7ff815b68943f08bef79f083c09e55ca8c18db8e25033034b12fa31f1d43dab4R48

fkrause98:
> I understand this is just a partial implementation of far call, but is it possible to add some tests with what we already have?

Sure, between the merges and the scope of this PR I lost track of testing, I'll be sure to add some more tests for this feature.

tests/integration_test.rs
gianbelinche: Perfect, I was thinking about tests running in parallel, but they don't

src/op_handlers/far_call.rs
fkrause98: You're right! Fixing it now

fkrause98:
With the new changes, I think it's sensible to ignore [the tests](https://github.com/lambdaclass/era_vm/pull/52/commits/72771093fde204fab7a02d5ab9600824762a9096#diff-7ff815b68943f08bef79f083c09e55ca8c18db8e25033034b12fa31f1d43dab4R779) for now until we have the deployer code ready.
Otherwise we would have to come up with a new decommit logic for the code, let's discuss this though.

jrchatruc:
Closed because these changes are included in #63

Issue #45, opened 2024-06-25
Add proper error handling


jrchatruc:
Remove all panics/unwraps/expects from the codebase and return appropriate results/errors instead.

Note that a lot of possible errors should result in execution reverts, meaning the current context/callframe is popped, all state changes to that context are reverted (except on OOG) and a false boolean value is returned to signify the revert.

PR #53, opened 2024-06-27
Near call Ret, Revert and Panic

gianbelinche:
Implementation of `Ret`, `Revert` and `Panic` opcodes for near calls


src/lib.rs
```diff
@@ -155,15 +158,27 @@ pub fn run(mut vm: VMState) -> (U256, VMState) {
...
+                        let should_break = _ok(&mut vm, &opcode);
+                        if should_break {
+                            break;
+                        }
                     }
-                }
+                    RetOpcode::Revert => {
+                        let should_break = _revert(&mut vm, &opcode);
+                        if should_break {
+                            panic!("Contract Reverted");
```
fkrause98, line 171: Since we're implementing the panic opcode, should still be a panic?
Maybe a Revert should be reported as an Err instead of panicking.

src/lib.rs
```diff
@@ -155,15 +158,27 @@ pub fn run(mut vm: VMState) -> (U256, VMState) {
...
+                        }
+                    }
+                    RetOpcode::Panic => {
+                        let should_break = _panic(&mut vm, &opcode);
+                        if should_break {
```
fkrause98, line 176: Same here, also, returning an Err, let's use properly test and report which Error the VM returns. This would also make the run function have to return a Result type, extending this PR a little bit, so let's discuss it a bit, but I think it's time we start returning Errs instead of panics. Starting from this PR.

src/op_handlers/near_call.rs
```diff
@@ -10,7 +10,7 @@ use crate::{opcode::Opcode, state::VMState};
 pub fn _near_call(vm: &mut VMState, opcode: &Opcode) {
     let abi_reg = vm.get_register(opcode.src0_index);
     let callee_address = opcode.imm0;
-    let _exception_handler = opcode.imm1; //TODO: Add exception handler to call frame
+    let exception_handler = opcode.imm1; //TODO: Add exception handler to call frame
```
fkrause98, line 13: Exception handlers will live in the same contract since this is a near call, right? Wouldn't this simply be a PC modification to then jump to that routine? Can't we implement this here?

src/op_handlers/near_call.rs
```diff
@@ -42,6 +42,7 @@ pub fn _near_call(vm: &mut VMState, opcode: &Opcode) {
         new_storage,
         callee_ergs,
         new_transient_storage,
+        exception_handler as u64,
```
fkrause98, line 45: If the spec specifies which bytes from the given U256 are used for the exception handler, let's explicitly cast it like that, else let's leave it as is for now.

src/state.rs
```diff
@@ -137,23 +137,17 @@ impl VMState {
         let new_context = Context::new(program_code, gas_stipend);
         self.running_contexts.push(new_context);
     }
-    pub fn pop_context(&mut self) {
-        self.running_contexts.pop();
+    pub fn pop_context(&mut self) -> Context {
+        self.running_contexts.pop().unwrap()
```
fkrause98, line 141: I think this is a good place to state what happened:
```suggestion
        self.running_contexts.pop().expect('Error: No running context')
```

tests/integration_test.rs
```diff
@@ -1702,7 +1691,105 @@ fn test_near_call_callee_uses_gas() {
...
+    let bin_path = make_bin_path_asm("near_call_revert_with_label");
+    let (result, _) = run_program(&bin_path);
+    assert_eq!(result, U256::from(7));
+}
+
+#[test]
+fn test_near_call_panics_with_label() {
+    let bin_path = make_bin_path_asm("near_call_panics_with_label");
+    let (result, _) = run_program(&bin_path);
+    assert_eq!(result, U256::from(7));
```
fkrause98, line 1792: Always nice to see something extended extensively, great work! 🚀 .

fkrause98 (changes requested):

src/lib.rs
gianbelinche: I left it as a panic in order to continue the same approach, I agree with you, this is being implemented on https://github.com/lambdaclass/era_vm/issues/45 , if you are ok, we can leave this one, and focus on error handling on that issue, since its a big change

src/lib.rs
gianbelinche: Same as above

src/op_handlers/near_call.rs
gianbelinche: I'm not sure what you mean, the exception handler is indeed what you say, a new PC where we want to jump, and that's whats being done

gianbelinche: I'll remove the TODO

src/op_handlers/near_call.rs
gianbelinche: The exception handler is not received in a register (U256) but from an immediate value (u16), so the conversion to u64 will never fail, it is only done since it needs to be used as the PC later

src/state.rs
gianbelinche: Sure!

tests/integration_test.rs
gianbelinche: Thanks!

fkrause98 (approved):

juan518munoz (approved):
LGTM

PR #54, opened 2024-06-28
Tracer Implementation

gianbelinche:
Closes https://github.com/lambdaclass/era_vm/issues/51
This PR implement the Tracer trait, right now it only has the function `before_execution`, in order to make it simpler, if we need more we can add them as we need it.

Two Tracers were implemented:
`StateSaverTracer` Saves the state of the vm before every execution
`PrintTracer` Makes Prints available on Yul and Solidity contracts

This PR depends on https://github.com/lambdaclass/era_vm/pull/53


src/lib.rs
```diff
@@ -83,23 +69,23 @@ pub fn program_from_file(bin_path: &str) -> Vec<U256> {
...
-/// Run a vm program from the given path using a custom state.
-/// Returns the value stored at storage with key 0 and the final vm state.
-pub fn run_program_with_custom_state(bin_path: &str, mut vm: VMState) -> (U256, VMState) {
-    let program = program_from_file(bin_path);
-    vm.load_program(program);
-    run(vm)
-}
-
-pub fn run(mut vm: VMState) -> (U256, VMState) {
+pub fn run(mut vm: VMState, tracers: &mut [Box<&mut dyn Tracer>]) -> (U256, VMState) {
```
fkrause98, line 82: Why not make Tracers a part of the VM?
I think it makes sense to do so, plus you don't need to pass around a vec/struct of Tracers around for when you want to run them. Plus we could have some extra logic for when to run them, and having a tracer as vm field, will make it easier to access the VM without having to fight with the borrow checker.

fkrause98 (changes requested):
Looks good, let's discuss where the tracers should live before approving.

src/lib.rs
gianbelinche: Good idea!

fkrause98 (approved):
LGTM

IAvecilla (approved):
LGTM!

PR #60, opened 2024-07-04
Error handling

juan518munoz:


src/bin/main.rs
Oppen, line 1: The check for `len() <= 1` was sufficient here, since it's `main`. You can remove the `unwrap()` in `args.get(1).unwrap()` simply by using `args[1]`. Do note that because of the previous `len()` check it will always succeed.

src/lib.rs
```diff
@@ -52,10 +53,11 @@ use zkevm_opcode_defs::{BinopOpcode, RetOpcode};
...
 /// Returns the value stored at storage with key 0 and the final vm state.
-pub fn program_from_file(bin_path: &str) -> Vec<U256> {
-    let program = std::fs::read(bin_path).unwrap();
-    let encoded = String::from_utf8(program.to_vec()).unwrap();
-    let bin = hex::decode(&encoded[2..]).unwrap();
+pub fn program_from_file(bin_path: &str) -> Result<Vec<U256>, EraVmError> {
+    let program = std::fs::read(bin_path)?;
+    let encoded =
+        String::from_utf8(program.to_vec()).map_err(|_| EraVmError::IncorrectBytecodeFormat)?;
+    let bin = hex::decode(&encoded[2..]).map_err(|_| EraVmError::IncorrectBytecodeFormat)?;
```
Oppen, line 60: We should probably check that `encoded[..2] == "0x"` as well.

src/lib.rs
```diff
@@ -52,10 +53,11 @@ use zkevm_opcode_defs::{BinopOpcode, RetOpcode};
...
 /// Run a vm program from the given path using a custom state.
 /// Returns the value stored at storage with key 0 and the final vm state.
-pub fn program_from_file(bin_path: &str) -> Vec<U256> {
-    let program = std::fs::read(bin_path).unwrap();
-    let encoded = String::from_utf8(program.to_vec()).unwrap();
-    let bin = hex::decode(&encoded[2..]).unwrap();
+pub fn program_from_file(bin_path: &str) -> Result<Vec<U256>, EraVmError> {
+    let program = std::fs::read(bin_path)?;
+    let encoded =
+        String::from_utf8(program.to_vec()).map_err(|_| EraVmError::IncorrectBytecodeFormat)?;
```
Oppen, line 59: Calling `to_vec()` is unnecessary. `std::fs::read` returns a `Vec<u8>` already.

src/lib.rs
```diff
@@ -65,124 +67,147 @@ pub fn program_from_file(bin_path: &str) -> Vec<U256> {
...
-                    Variant::Shift(ShiftOpcode::Shl) => _shl(&mut vm, &opcode),
-                    Variant::Shift(ShiftOpcode::Shr) => _shr(&mut vm, &opcode),
-                    Variant::Shift(ShiftOpcode::Rol) => _rol(&mut vm, &opcode),
-                    Variant::Shift(ShiftOpcode::Ror) => _ror(&mut vm, &opcode),
+                    Variant::Shift(ShiftOpcode::Shl) => _shl(&mut vm, &opcode)?,
+                    Variant::Shift(ShiftOpcode::Shr) => _shr(&mut vm, &opcode)?,
+                    Variant::Shift(ShiftOpcode::Rol) => _rol(&mut vm, &opcode)?,
+                    Variant::Shift(ShiftOpcode::Ror) => _ror(&mut vm, &opcode)?,
                     _ => unreachable!(),
                 },
```
Oppen, line 142: We could remove the `unreachable!()` by taking the shift variant from the outer `match`:
```rust
Variant::Shift(shift_variant) => match shift_variant {
    ShiftOpcode::Shl => _shl(&mut vm, &opcode)?,
    ShiftOpcode::Shr => _shr(&mut vm, &opcode)?,
    ShiftOpcode::Rol => _rol(&mut vm, &opcode)?,
    ShiftOpcode::Ror => _ror(&mut vm, &opcode)?,
},
```
This works because it makes the inner `match` exhaustive. And, while the original code is correct, this way we can get help from the compiler if we miss one of the inner arms, due to the removal of the catch-all `_` arm.

src/lib.rs
```diff
@@ -65,124 +67,147 @@ pub fn program_from_file(bin_path: &str) -> Vec<U256> {
...
-                    UMAOpcode::AuxHeapRead => _aux_heap_read(&mut vm, &opcode),
-                    UMAOpcode::AuxHeapWrite => _aux_heap_write(&mut vm, &opcode),
-                    UMAOpcode::FatPointerRead => _fat_pointer_read(&mut vm, &opcode),
+                    UMAOpcode::HeapRead => _heap_read(&mut vm, &opcode)?,
+                    UMAOpcode::HeapWrite => _heap_write(&mut vm, &opcode)?,
+                    UMAOpcode::AuxHeapRead => _aux_heap_read(&mut vm, &opcode)?,
+                    UMAOpcode::AuxHeapWrite => _aux_heap_write(&mut vm, &opcode)?,
+                    UMAOpcode::FatPointerRead => _fat_pointer_read(&mut vm, &opcode)?,
                     UMAOpcode::StaticMemoryRead => todo!(),
                     UMAOpcode::StaticMemoryWrite => todo!(),
```
Oppen, line 205: If we want to avoid crashes we may want to add an `_unimplemented()` helper that matches the signature of other opcodes and reverts and does proper error reporting.

src/op_handlers/aux_heap_read.rs
```diff
@@ -2,24 +2,29 @@ use u256::U256;
...
+        ));
     }
 
     if src0.value > U256::from(MAX_OFFSET_TO_DEREF_LOW_U32) {
-        panic!("Address too large for heap_read");
+        return Err(EraVmError::OperandError(
+            "Address too large for heap_read".to_string(),
+        ));
     }
     let addr = src0.value.low_u32();
```
Oppen, line 22: It makes a ton more sense that `low_u32()` returns a `Result`. Then, it can have a proper error variant for out of bounds.

Oppen: Or, rather, have an `impl TryFrom<U256> for u32`.

src/op_handlers/aux_heap_write.rs
```diff
@@ -2,24 +2,29 @@ use u256::U256;
...
+        ));
     }
 
     if src0.value > U256::from(MAX_OFFSET_TO_DEREF_LOW_U32) {
-        panic!("Address too large for heap_write");
+        return Err(EraVmError::OperandError(
+            "Address too large for heap_write".to_string(),
+        ));
     }
     let addr = src0.value.low_u32();
```
Oppen, line 22: Same here.

src/op_handlers/heap_read.rs
```diff
@@ -2,24 +2,29 @@ use u256::U256;
...
+        ));
     }
 
     if src0.value > U256::from(MAX_OFFSET_TO_DEREF_LOW_U32) {
-        panic!("Address too large for heap_read");
+        return Err(EraVmError::OperandError(
+            "Address too large for heap_read".to_string(),
+        ));
     }
     let addr = src0.value.low_u32();
```
Oppen, line 22: Ditto.

src/op_handlers/heap_write.rs
```diff
@@ -2,24 +2,29 @@ use u256::U256;
...
+        ));
     }
 
     if src0.value > U256::from(MAX_OFFSET_TO_DEREF_LOW_U32) {
-        panic!("Address too large for heap_write");
+        return Err(EraVmError::OperandError(
+            "Address too large for heap_write".to_string(),
+        ));
     }
     let addr = src0.value.low_u32();
```
Oppen, line 22: Ditto.

src/ptr_operator.rs
```diff
@@ -2,33 +2,41 @@ use zkevm_opcode_defs::MAX_OFFSET_FOR_ADD_SUB;
...
+pub fn ptr_operands_read(
+    vm: &mut VMState,
+    opcode: &Opcode,
+) -> Result<(FatPointer, u32, TaggedValue), EraVmError> {
+    let (src0, src1) = address_operands_read(vm, opcode)?;
 
     if !src0.is_pointer || src1.is_pointer {
-        panic!("Invalid operands for {:?}", opcode.variant);
+        return Err(EraVmError::OperandError(
+            "Invalid operands for ptr_shrink".to_string(),
```
Oppen, line 19: Maybe `OperandError` could contain the opcode variant instead.

Oppen:
I see most of the errors are stringly typed. Is this a temporary shortcut so we have some error handling today rather than perfect error handling in a month or is it meant to be permanent?

src/bin/main.rs
juan518munoz: It looks like the most recent changes to this section weren't pushed when you reviewed, it now looks like this:
```rs
fn main() {
    let args: Vec<String> = env::args().collect();
    let bin_path = match args.get(1) {
        Some(path) => path,
        None => {
            println!("Pass a program to run");
            exit(1);
        }
    };

    let vm = VMStateBuilder::default().build();
    let output = run_program(bin_path, vm, &mut []);
    println!("RESULT: {:?}", output.storage_zero);
}
```
What do you think of it?

src/lib.rs
juan518munoz: Done!

src/lib.rs
juan518munoz: Correct.

src/lib.rs
```diff
@@ -65,124 +67,147 @@ pub fn program_from_file(bin_path: &str) -> Vec<U256> {
...
+            reverted: false,
+            reason: None,
+        },
+        Err(e) => Output {
+            storage_zero: U256::zero(),
+            vm_state: VMState::default(),
+            reverted: true,
+            reason: Some(e),
+        },
+    }
```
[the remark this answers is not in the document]
juan518munoz: `Output` can not implement `Default`, as `Err` needs to be assinged a specific variable:
```rs
       ...
        Err(e) => Output {
            storage_zero: U256::zero(),
            vm_state: VMState::default(),
            reverted: true,
            reason: Some(e), //  <-- prohibits Default
        },
 ```
 Am I missing something?

File: src/lib.rs (hunk 3 above)
Comment (juan518munoz) [reply src/lib.rs]: Correct, it has been applied in one of the latests commits.

File: src/lib.rs (hunk 4 above)
Comment (juan518munoz) [reply src/lib.rs:198]: I'm not sure if it is a needed change, as we are running the vm only with test which only use instructions already implemented. 

It may be useful tho for when we get further in development of the virtual machine, and run programs that are not from our tests.

File: src/op_handlers/aux_heap_read.rs (hunk 5 above)
Comment (juan518munoz) [reply src/op_handlers/aux_heap_read.rs:22]: Adding `TryFrom` would be nice, but `U256` is an external type, so it's not possible to implement it.

Same goes for the method `low_u32()`, it's not possible to modify it.

I can't think of any other way of doing this than what is currently done (manual check, followed by assignment to variable).

File: src/op_handlers/aux_heap_write.rs (hunk 6 above)
Comment (juan518munoz) [reply src/op_handlers/aux_heap_write.rs:22]: Same answer as above.

File: src/op_handlers/heap_read.rs (hunk 7 above)
Comment (juan518munoz) [reply src/op_handlers/heap_read.rs:22]: Same as above.

File: src/op_handlers/heap_write.rs (hunk 8 above)
Comment (juan518munoz) [reply src/op_handlers/heap_write.rs:22]: Again, same as above.

File: src/ptr_operator.rs (hunk 9 above)
Comment (juan518munoz) [reply src/ptr_operator.rs:19]: You are right, adding it.

**Comment by juan518munoz** [comment]
I believe that we should at least replace the _stringly_ errors with `Enums` before pushing this PR to `main`, do you suggest anything else?

File: src/lib.rs (hunk 10 above)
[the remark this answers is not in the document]
Comment (Oppen) [reply src/lib.rs:92]: No, I missed that it was assigning the error 🤦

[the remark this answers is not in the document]
Comment (Oppen) [reply src/lib.rs:92]: That said, implementing `from` for the relevant types (`(U256, VMState)` and `EraVmError`) enables the following:
```rust
run_program_with_error(bin_path, vm, tracers)
    .map(Output::from)
    .ok_or_else(Output::from)
```
It may or may not require using the turbofish to specify types, but my guess is the compiler can infer them.

src/op_handlers/aux_heap_read.rs
Oppen: Sad. We could use a function or a local trait that mimics `TryFrom`'s behavior. If the trait is local to our crate then the orphan rule is satisfied.

Oppen:
I think that's the way to go. The question was mostly if we intended to do that or not.

src/op_handlers/aux_heap_read.rs
juan518munoz: Following this, `U256` actually implements `try_into()` `u32`, but it doesn't fit our case, because we are not trying to check wheter the value is lower or equal than `u32::MAX`, but rather `MAX_OFFSET_TO_DEREF_LOW_U32`, which is smaller.

So we could implement a trait like this:
```rs
pub trait IntoU32Address {
    fn into_u32_address(self, opcode: Variant) -> Result<u32, OperandError>;
}

impl IntoU32Address for U256 {
    fn into_u32_address(self, opcode: Variant) -> Result<u32, OperandError> {
        if self > U256::from(MAX_OFFSET_TO_DEREF_LOW_U32) {
            return Err(OperandError::InvalidSrcAddress(opcode));
        }
        Ok(self.low_u32())
    }
}
```

And use it like this:

```rs
pub fn aux_heap_read(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmError> {
    let (src0, _) = address_operands_read(vm, opcode)?;
    if src0.is_pointer {
        return Err(OperandError::InvalidSrcPointer(opcode.variant).into());
    }

    let addr = src0.value.into_u32_address(opcode.variant)?; // <-- HERE
    let gas_cost = vm.current_frame_mut()?.aux_heap.expand_memory(addr + 32);
    vm.current_frame_mut()?.gas_left -= gas_cost;

    let value = vm.current_frame_mut()?.aux_heap.read(addr);
    vm.set_register(opcode.dst0_index, TaggedValue::new_raw_integer(value));

    if opcode.alters_vm_flags {
        // This flag is set if .inc is present
        vm.set_register(
            opcode.dst1_index,
            TaggedValue::new_raw_integer(U256::from(addr + 32)),
        );
    }
    Ok(())
}
```

Oppen:
A general note (I'm not sure if it's still applies): when an error path allocates, for example when `.to_string()` is called, `.ok_or_else` is preferred over `.ok_or`, since the latter takes a value and thus makes allocation unconditional, making the "happy path" slower, while the former receives a closure that is only called when the result is an error, only allocating in the error path.

src/address_operands.rs
```diff
@@ -39,46 +47,46 @@ pub fn address_operands_read(vm: &mut VMState, opcode: &Opcode) -> (TaggedValue,
                     // stack-=[src0 + offset] + src1
                     let (src0, src1) = reg_and_imm_read(vm, opcode);
                     let res = *vm
-                        .current_frame_mut()
+                        .current_frame_mut()?
                         .stack
-                        .get_with_offset(src0.value.as_usize());
-                    vm.current_frame_mut().stack.pop(src0.value);
+                        .get_with_offset(src0.value.as_usize())?;
```
fkrause98, line 52: Let's use try_from here instead, since this implementation panics without change of returning an Err.

src/address_operands.rs
```diff
@@ -39,46 +47,46 @@ pub fn address_operands_read(vm: &mut VMState, opcode: &Opcode) -> (TaggedValue,
...
                 }
                 ImmMemHandlerFlags::UseAbsoluteOnStack => {
                     // stack=[src0 + offset] + src1
                     let (src0, src1) = reg_and_imm_read(vm, opcode);
                     let res = vm
-                        .current_frame_mut()
+                        .current_frame_mut()?
                         .stack
-                        .get_absolute(src0.value.as_usize());
+                        .get_absolute(src0.value.as_usize())?;
```
fkrause98, line 72: Same as above

src/address_operands.rs
```diff
@@ -39,46 +47,46 @@ pub fn address_operands_read(vm: &mut VMState, opcode: &Opcode) -> (TaggedValue,
...
 
                     (*res, src1)
                 }
                 ImmMemHandlerFlags::UseImm16Only => only_imm16_read(vm, opcode),
                 ImmMemHandlerFlags::UseCodePage => {
                     let (src0, src1) = reg_and_imm_read(vm, opcode);
 
-                    let res = vm.current_frame_mut().code_page[src0.value.as_usize()];
+                    let res = vm.current_frame_mut()?.code_page[src0.value.as_usize()];
```
fkrause98, line 80: Ditto, also let's use [.get](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.get) here.

src/address_operands.rs
```diff
@@ -146,7 +162,7 @@ fn address_operands(vm: &mut VMState, opcode: &Opcode, res: (TaggedValue, Option
                 only_reg_write(vm, opcode, OutputOperandPosition::First, res.0);
             }
             RegOrImmFlags::UseImm16Only => {
-                panic!("dest cannot be imm16 only");
+                return Err(OperandError::InvalidDestImm16Only(opcode.variant).into());
```
fkrause98, line 165: Can we double check this into does not panic?

src/address_operands.rs
```diff
@@ -157,34 +173,37 @@ fn address_operands(vm: &mut VMState, opcode: &Opcode, res: (TaggedValue, Option
...
+                    vm.current_frame_mut()?.stack.store_with_offset(1, res.0)?;
                 }
                 ImmMemHandlerFlags::UseStackWithOffset => {
                     // stack[src0 + offset] + src1
                     let src0 = reg_and_imm_write(vm, OutputOperandPosition::First, opcode);
-                    vm.current_frame_mut()
+                    vm.current_frame_mut()?
                         .stack
-                        .store_with_offset(src0.value.as_usize(), res.0);
+                        .store_with_offset(src0.value.as_usize(), res.0)?;
```
fkrause98, line 186: Same as above

src/output.rs
```diff
@@ -0,0 +1,10 @@
+use u256::U256;
+
+use crate::{eravm_error::EraVmError, state::VMState};
+
+pub struct Output {
```
fkrause98, line 5: Nice

src/tracers/print_tracer.rs
```diff
@@ -24,13 +24,21 @@ impl Tracer for PrintTracer {
 
         if matches!(opcode_variant, ZKOpcode::UMA(UMAOpcode::HeapWrite)) {
             let mut new_vm = vm.clone();
-            let (src0, src1) = address_operands_read(&mut new_vm, opcode);
+            let (src0, src1) = address_operands_read(&mut new_vm, opcode).unwrap();
```
fkrause98, line 27: Let's avoid this unwrap

src/tracers/print_tracer.rs
```diff
@@ -24,13 +24,21 @@ impl Tracer for PrintTracer {
...
             let value = src1.value;
             if value == debug_magic {
                 let fat_ptr = FatPointer::decode(src0.value);
                 if fat_ptr.offset == DEBUG_SLOT {
-                    let how_to_print_value = new_vm.current_frame_mut().heap.read(DEBUG_SLOT + 32);
-                    let value_to_print = new_vm.current_frame_mut().heap.read(DEBUG_SLOT + 64);
+                    let how_to_print_value = new_vm
+                        .current_frame_mut()
+                        .unwrap()
```
fkrause98, line 34: Ditto

src/tracers/print_tracer.rs
```diff
@@ -24,13 +24,21 @@ impl Tracer for PrintTracer {
...
+                        .heap
+                        .read(DEBUG_SLOT + 32);
+                    let value_to_print = new_vm
+                        .current_frame_mut()
+                        .unwrap()
```
fkrause98, line 39: Ditto

fkrause98 (approved):
Looks good, let's address the comments I left and we're done.

src/address_operands.rs
juan518munoz: Done!

src/tracers/print_tracer.rs
juan518munoz: Removed all `unwrap()`s related to `EraVmError` and only left the ones related to `U256::from_str_radix()` that we know won't fail.

src/op_handlers/revert.rs
```diff
@@ -1,49 +1,75 @@
...
 
-pub fn revert_out_of_gas(vm: &mut VMState) {
+fn revert_near_call(vm: &mut VMState) -> Result<(), EraVmError> {
+    let previous_frame = vm.pop_frame()?;
+
+    let current_frame = vm.current_frame_mut()?;
+    // current_frame.stack = previous_frame.stack;
+    // current_frame.heap = previous_frame.heap;
+    // current_frame.aux_heap = previous_frame.aux_heap;
+    current_frame.pc = previous_frame.exception_handler - 1; // To account for the +1 later
```
gianbelinche, line 40: Don't we still want to preserve the changes on stack and heaps even if it reverts?

tests/integration_test.rs
```diff
@@ -2071,65 +2299,59 @@ fn test_all_modifiers() {
...
 }
 
-#[test]
-fn test_near_call_panics_with_label() {
-    let bin_path = make_bin_path_asm("near_call_panics_with_label");
-    let vm = VMStateBuilder::default().build();
-    let (result, _) = run_program(&bin_path, vm, &mut []);
-    assert_eq!(result, U256::from(7));
-}
-
```
gianbelinche, line 2087: Why was this test removed?

gianbelinche (changes requested):
Maybe add a test with a near call were the call reverts due to an error?

tests/integration_test.rs
juan518munoz: It was not removed, only moved to another place (line 2251).

src/op_handlers/revert.rs
juan518munoz: You are correct. I commented them out during testing, and forgot to re-add them.

juan518munoz:
> Maybe add a test with a near call were the call reverts due to an error?

What would be the scope of this test? We already have tests like `near_call_panic` that revert on a near call.

juan518munoz:
> Maybe add a test with a near call were the call reverts due to an error?

Added `test_near_call_error_revert()`.

gianbelinche (approved):

gianbelinche:
LGTM!

PR #78, opened 2024-07-17
Fix smod and sdiv

gianbelinche:
This PR fixes the smod and sdiv tests

On the formal specification, the div opcode states:
```
Flags are computed as follows:
◦ LT_OF is cleared;
◦ EQ is set if the quotient is not zero;
◦ GT is set if the reminder is not zero.
```

However, on both the vm2 and zk-evm, EQ and GT are set if quotient and reminder are 0, respectively.

### DIFF
--- a/src/op_handlers/div.rs
+++ b/src/op_handlers/div.rs
@@ -8,14 +8,27 @@ use crate::{opcode::Opcode, state::VMState};
 pub fn div(vm: &mut VMState, opcode: &Opcode) {
     let (src0_t, src1_t) = address_operands_read(vm, opcode);
     let (src0, src1) = (src0_t.value, src1_t.value);
-    let (quotient, remainder) = src0.div_mod(src1);
-    if opcode.alters_vm_flags {
-        // Lt overflow is cleared
-        vm.flag_lt_of = false;
-        // Eq is set if quotient is zero
-        vm.flag_eq = quotient.is_zero();
-        // Gt is set if the remainder is zero
-        vm.flag_gt = remainder.is_zero();
+    let mut quotient = U256::zero();
+    let mut remainder = U256::zero();
+    if src1.is_zero() {
+        if opcode.alters_vm_flags {
+            // Lt overflow is set
+            vm.flag_lt_of = true;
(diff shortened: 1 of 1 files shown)


juan518munoz:
It looks like we are not taking into account the case `in2 == 0`, that changes the values of the outputs (both are set to `0`) and how the flags are set. It's not that related to this PR, but might be a good oportunity to add such a small correction here.

[Source](https://matter-labs.github.io/eravm-spec/spec.html#:~:text=%E2%86%92%20flags_tsmallstep%20%3A%3D-,OpDiv,-Abstract%20Syntax)

src/op_handlers/div.rs
```diff
@@ -1,3 +1,6 @@
+
+use u256::U256;
```
jorbush, line 2: Is this import necessary? I don't see where it is used in the new changes.

jorbush (changes requested):

src/op_handlers/div.rs
gianbelinche: It wasn't, now It is :)

juan518munoz (approved):
LGTM

src/op_handlers/div.rs
```diff
@@ -6,14 +8,27 @@ use crate::{opcode::Opcode, state::VMState};
...
+        if opcode.alters_vm_flags {
+            // Lt overflow is set
+            vm.flag_lt_of = true;
+            // Eq is set if resultlow is 0, in this case its always 0
+            vm.flag_eq = true;
+            // Gt is set if LT_OF and EQ are cleared, they are not
+            vm.flag_gt = false;
+        }
+    } else {
+        (quotient, remainder) = src0.div_mod(src1);
```
jorbush, line 23: Maybe it's a little fussy but i think that if only these variables are used inside the condition, this declaration should be inside it, because if the `allter_flags` is not enabled, an extra operation is done that is not used.

jorbush (changes requested):

jorbush (approved):
LGTM

jrchatruc (approved):

PR #81, opened 2024-07-18
Fix return tests

gianbelinche:
This PR fixes the tests on return.sol

For that I added some logic regarding far call reverts, specifically when an opcode inside a far call throws an error, however it is just the necessary stuff to make this tests pass, its not complete, and there could even be some unnecessary things.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -207,25 +207,37 @@ pub fn run(
                 // This is only to keep the context for tests
                 Variant::Ret(ret_variant) => match ret_variant {
                     RetOpcode::Ok => {
-                        let should_break = ok(&mut vm, &opcode)?;
-                        if should_break {
-                            break;
+                        let should_break = ok(&mut vm, &opcode);
+                        if should_break.is_err() {
+                            Err(should_break.unwrap_err())
+                        } else {
+                            if should_break.unwrap() {
+                                break;
+                            }
+                            Ok(())
                         }
-                        Ok(())
                     }
                     RetOpcode::Revert => {
-                        let should_break = revert(&mut vm, &opcode)?;
-                        if should_break {
-                            return Ok((ExecutionOutput::Revert(vec![]), vm));
+                        let should_break = revert(&mut vm, &opcode);
+                        if should_break.is_err() {
+                            Err(should_break.unwrap_err())
+                        } else {
+                            if should_break.unwrap() {
+                                return Ok((ExecutionOutput::Revert(vec![]), vm));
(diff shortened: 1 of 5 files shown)


jorbush (approved):
LGTM

src/lib.rs
```diff
@@ -207,25 +207,37 @@ pub fn run(
...
+                            Err(should_break.unwrap_err())
+                        } else {
+                            if should_break.unwrap() {
+                                return Ok((ExecutionOutput::Panic, vm));
+                            }
+                            Ok(())
                         }
-                        Ok(())
                     }
                 },
```
juan518munoz, line 242: Simplify by `match`ing:
```rs
                Variant::Ret(ret_variant) => match ret_variant {
                    RetOpcode::Ok => match ok(&mut vm, &opcode) {
                        Ok(should_break) => {
                            if should_break {
                                break;
                            }
                            Ok(())
                        }
                        Err(e) => Err(e),
                    },
                    RetOpcode::Revert => match revert(&mut vm, &opcode) {
                        Ok(should_break) => {
                            if should_break {
                                return Ok((ExecutionOutput::Revert(vec![]), vm));
                            }
                            Ok(())
                        }
                        Err(e) => Err(e),
                    },
                    RetOpcode::Panic => match panic(&mut vm, &opcode) {
                        Ok(should_break) => {
                            if should_break {
                                return Ok((ExecutionOutput::Panic, vm));
                            }
                            Ok(())
                        }
                        Err(e) => Err(e),
                    },
                },
```

juan518munoz (changes requested):

src/lib.rs
juan518munoz: Even simpler (maybe not that legible?):
```rs
                Variant::Ret(ret_variant) => {
                    let result = match ret_variant {
                        RetOpcode::Ok => ok(&mut vm, &opcode),
                        RetOpcode::Revert => revert(&mut vm, &opcode),
                        RetOpcode::Panic => panic(&mut vm, &opcode),
                    };
                    match result {
                        Ok(should_break) => {
                            if should_break {
                                break;
                            }
                            Ok(())
                        }
                        Err(e) => Err(e),
                    }
                }
```

gianbelinche: The second one would not work, since the return values are different, I'll do the first one, thanks!

juan518munoz: You are right, missed it!

juan518munoz (approved):
LGTM

PR #83, opened 2024-07-19
Fix log tests

gianbelinche:
This PR fixes log0 to log4 tests

### DIFF
--- a/src/call_frame.rs
+++ b/src/call_frame.rs
@@ -47,6 +47,7 @@ impl Context {
         heap_id: u32,
         aux_heap_id: u32,
         calldata_heap_id: u32,
+        exception_handler: u64
     ) -> Self {
         Self {
             frame: CallFrame::new_far_call_frame(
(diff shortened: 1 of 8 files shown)


jorbush (approved):
LGTM!

juan518munoz:
Not part of this PR, but related, there are two other places in our codebase where `get_forward_memory_pointer ` is called

https://github.com/lambdaclass/era_vm/blob/323a83b902373c24becf335e500b1b0a80e677f2/src/op_handlers/ok.rs#L29-L30
https://github.com/lambdaclass/era_vm/blob/323a83b902373c24becf335e500b1b0a80e677f2/src/op_handlers/ok.rs#L40-L41

In this case and neither any of those other cases the `None` case is handled, it might be better to just propagate the error the function might throw.

Change:
```rs
pub fn get_forward_memory_pointer(
    source: U256,
    vm: &mut VMState,
    is_pointer: bool,
) -> Result<Option<FatPointer>, EraVmError>
```

To:
```rs
pub fn get_forward_memory_pointer(
    source: U256,
    vm: &mut VMState,
    is_pointer: bool,
) -> Result<FatPointer, EraVmError>
```

juan518munoz (approved):
LGTM

PR #85, opened 2024-07-22
Remove `--disable-deployer` flag from CI tests

jrchatruc:


gianbelinche:
Done with https://github.com/lambdaclass/era_vm/pull/91

PR #89, opened 2024-07-23
Add detailed doc on the VM

jrchatruc:


docs/evm_vs_eravm.md
```diff
@@ -0,0 +1,309 @@
...
+
+```
+PUSH1 0x80
+PUSH1 0x40
+MSTORE
+CALLVALUE
+DUP1
+ISZERO
+PUSH1 0xE
+JUMPI
```
Oppen, line 27: This assembly is weird, why is there no addition? Is the code for the function or a call site?

gianbelinche: This is not the complete assembly, almost all compiled solidity contracts start this way, it is basically checking which function to execute

Oppen: OK, but is it not supposed to be the assembly for the sample snippet?

jrchatruc: @Oppen Ideally it would be, the problem is it's really hard to isolate just the code for the snippet. `solc` generates the entire code for the contract function dispatching, etc.

The idea here was more to show how different both assemblies are (one is a super simple stack based one, the other one a complex register based one) at a high level

PR #90, opened 2024-07-23
Era VM integration with `zksync-era`

IAvecilla:
This PR modifies the era vm to be compatible with zksync-era, for that a new Storage implementation was developed


src/store.rs
```diff
@@ -15,6 +15,7 @@ pub trait Storage: Debug {
     fn add_contract(&mut self, hash: U256, code: Vec<U256>) -> Result<(), StorageError>;
     fn storage_read(&self, key: StorageKey) -> Result<Option<U256>, StorageError>;
     fn storage_write(&mut self, key: StorageKey, value: U256) -> Result<(), StorageError>;
+    fn get_state_storage(&self) -> &HashMap<StorageKey, U256>;
```
Oppen, line 18: I believe the idea of the abstraction as it is made is that we ask for the value we need by calling `storage_read(key)` instead of returning the `HashMap`.

src/store.rs
```diff
@@ -8,130 +9,159 @@ use zkevm_opcode_defs::{
...
-#[derive(Error, Debug)]
-pub enum StorageError {
-    #[error("Key not present in storage")]
-    KeyNotPresent,
-    #[error("Error writing to storage")]
-    WriteError,
-    #[error("Error reading from storage")]
-    ReadError,
+#[derive(Debug, Clone)]
+pub struct InitialStorageMemory {
```
fkrause98, line 27: Let's add some comments  here on what this trait represents and why would you implement it.

src/store.rs
```diff
@@ -8,130 +9,159 @@ use zkevm_opcode_defs::{
...
-        InMemory {
-            state_storage,
-            contract_storage,
-        }
-    }
+#[derive(Debug)]
+pub struct StateStorage {
+    pub storage_changes: HashMap<StorageKey, U256>,
+    pub initial_storage: Rc<RefCell<dyn InitialStorage>>,
+    l1_to_l2_logs: Vec<L2ToL1Log>,
```
fkrause98, line 55: The field is` l1_to_l2_logs` but the type is L2ToL1Log, so which one is it then?

src/store.rs
```diff
@@ -8,130 +9,159 @@ use zkevm_opcode_defs::{
...
 
-    fn storage_read(&self, key: StorageKey) -> Result<Option<U256>, StorageError> {
-        Ok(self.state_storage.get(&key).copied())
+    pub fn storage_read(&self, key: StorageKey) -> Result<Option<U256>, StorageError> {
+        let mut value = self.storage_changes.get(&key).copied();
+        if value.is_none() {
+            value = self.initial_storage.borrow().storage_read(key)?;
+        }
+        Ok(value)
     }
```
fkrause98, line 85: Nit-pick: an optional style suggestion, up to you if you want to apply this change 😄 .
```suggestion
        pub fn storage_read(&self, key: StorageKey) -> Result<Option<U256>, StorageError> {
        match self.storage_changes.get(&key) {
            None => self.initial_storage.borrow().storage_read(key),
            value => Ok(value.copied()),
        }
    }
```

src/op_handlers/log.rs
```diff
@@ -3,57 +3,78 @@ use u256::U256;
...
 }
+
+pub fn add_l2_to_l1_message(
+    vm_state: &mut VMState,
+    opcode: &Opcode,
+    state_storage: &mut StateStorage,
+) -> Result<(), EraVmError> {
+    let key = vm_state.get_register(opcode.src0_index).value;
+    let value = vm_state.get_register(opcode.src1_index).value;
+    let is_service = opcode.imm1 == 1;
```
MarcosNicolau, line 70: I believe this should be `imm0`

src/store.rs
```diff
@@ -8,130 +9,159 @@ use zkevm_opcode_defs::{
...
 
-    fn get_all_keys(&self) -> Vec<StorageKey> {
-        self.state_storage.keys().copied().collect()
+    pub fn rollback(&mut self, snapshot: &SnapShot) {
+        let keys = snapshot.storage_changes.keys();
+        for key in keys {
+            let value = snapshot.storage_changes.get(key);
+            if let Some(value) = value {
+                self.storage_write(*key, *value).unwrap();
+            }
```
fkrause98, line 109: Another (optional) style suggestion:
```suggestion
                 snapshot
                .storage_changes
                .get(key)
                .map(|value| self.storage_write(*key, *value));
```

src/store.rs
```diff
@@ -142,142 +172,5 @@ pub fn initial_decommit(storage: &mut dyn Storage, address: H160) -> Vec<U256> {
...
-
-pub fn encode_contract(contract: Vec<U256>) -> Vec<u8> {
-    let mut encoded: Vec<u8> = vec![];
-    for word in contract {
-        let encoded_word = encode(&word);
-        encoded.extend_from_slice(&encoded_word);
-    }
-
-    encoded
+    contract_storage.decommit(code_key).unwrap().unwrap()
```
fkrause98, line 175: I get that the idea here is that if the initial decommit fails, then the execution cannot continue, but I'd prefer to avoid unwraps here, so lets either:
A - Return a Result.
B - Replace unwrap with expect and an error description.

Although I would prefer option A since we don't forcefully panic on whoever calls this function.

fkrause98 (changes requested):
Hey @gianbelinche great work! Looks good, let's discus some things first before merging though.

MarcosNicolau (changes requested):

src/state.rs
```diff
@@ -151,18 +153,23 @@ pub struct VMState {
...
+    #[allow(clippy::too_many_arguments)]
     pub fn new(
         program_code: Vec<U256>,
         calldata: Vec<u8>,
         contract_address: H160,
         caller: H160,
         context_u128: u128,
         default_aa_code_hash: [u8; 32],
+        hook_address: u32,
+        use_hooks: bool,
```
Oppen, line 172: Maybe these could be an option?
```suggestion
        hook_address: Option<u32>,
```
`None` would imply you don't use hooks.

gianbelinche: I feel that would make the code uglier when checking if it needs to use hooks
Instead of doing `if use_hooks ...` it would have to do `if hook_address.is_some()` or something like
`if let Some(address) = hook_address ...`

MarcosNicolau (approved):
LGTM 👍

MarcosNicolau:
Doesn't this address issue #132?

gianbelinche:
> Btw, doesn't this address issue #132?

Yes

fkrause98 (approved):

src/state.rs
Oppen: It's uglier but it's less error prone, since you only have access to the address if you already checked you're using hooks.

PR #99, opened 2024-07-24
Fix immutable tests

gianbelinche:
This PR fixes the tests under `solidity/simple/immutable`

### DIFF
--- a/src/address_operands.rs
+++ b/src/address_operands.rs
@@ -51,7 +51,7 @@ pub fn address_operands_read(
                         .current_context()?
                         .stack
                         .get_with_offset(src0.value.as_usize(), sp)?;
-                    vm.current_frame_mut()?.sp -= src0.value.low_u64();
+                    vm.current_frame_mut()?.sp -= src0.value.low_u32() as u16;
                     (res, src1)
                 }
                 ImmMemHandlerFlags::UseStackWithOffset => {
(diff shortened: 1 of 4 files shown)


jorbush (approved):
Nice work! 
Also fixes: 
- `tests/solidity/simple/expression`
- `tests/solidity/simple/gas_value`

src/address_operands.rs
```diff
@@ -176,11 +176,7 @@ fn address_operands(
                 ImmMemHandlerFlags::UseStackWithPushPop => {
                     // stack+=[src0 + offset] + src1
                     let src0 = reg_and_imm_write(vm, OutputOperandPosition::First, opcode);
-                    vm.current_frame_mut()?.sp += src0.value.low_u64() + 1;
-                    let sp = vm.current_frame()?.sp;
-                    vm.current_context_mut()?
-                        .stack
-                        .store_with_offset(1, res.0, sp)?;
+                    vm.current_frame_mut()?.sp += src0.value.low_u64();
```
Oppen, line 179: Stacks are 16-bits addressed, so I'm not sure `low_u64`  is enough bits dropped.

src/op_handlers/far_call.rs
```diff
@@ -262,7 +262,6 @@ pub fn far_call(
 
             // set calldata pointer
             vm.set_register(1, TaggedValue::new_pointer(forward_memory.encode()));
-            vm.current_context_mut()?.caller = address_from_u256(&vm.get_register(15).value);
```
Oppen, line 265: I don't exactly understand this bit. Is this due to static mode being in effect?

gianbelinche: No, in mimic calls the caller is set to the value of register 15

On the formal spec:
```
Mimic call sets:
this_address <- destination address;
msg_sender <- value of r15;
context <- value of context register gs_context_u128
```

src/address_operands.rs
gianbelinche: Done

juan518munoz (approved):
LGTM

Oppen (approved):

PR #100, opened 2024-07-24
Tracer documentation

gianbelinche:


docs/general.md
```diff
@@ -34,7 +34,66 @@ function sendMoney(address payable to) public payable {
...
+```
+pub trait Tracer {
+    fn before_execution(&mut self, _opcode: &Opcode, _vm: &mut VMState) -> Result<(), EraVmError>;
+}
+```
+
+The `before_execution` function will be called on every loop just before the opcode execution.
+Right know that is the only function the trait has, in the future more may be added as needed, like `before_decoding`, `after_decoding` or `after_execution`
+
+An important Tracer is what whe call the `PrintTracer`, with it we can print stuff on solidity contracts.
```
Oppen, line 47:
```suggestion
An important Tracer is what we call the `PrintTracer`, with it we can print stuff on solidity contracts.
```

docs/general.md
```diff
@@ -34,7 +34,66 @@ function sendMoney(address payable to) public payable {
...
-How tracer works, how the PrintTracer works, how to add prints when running contracts.
+A `Tracer` should comply with the following trait
+```
+pub trait Tracer {
+    fn before_execution(&mut self, _opcode: &Opcode, _vm: &mut VMState) -> Result<(), EraVmError>;
+}
+```
+
+The `before_execution` function will be called on every loop just before the opcode execution.
+Right know that is the only function the trait has, in the future more may be added as needed, like `before_decoding`, `after_decoding` or `after_execution`
```
Oppen, line 45:
```suggestion
Right now that is the only function the trait has, in the future more may be added as needed, like `before_decoding`, `after_decoding` or `after_execution`
```

docs/general.md
```diff
@@ -34,7 +34,66 @@ function sendMoney(address payable to) public payable {
...
+        printItNum(result);
+        return result;
+    }
+}
+```
+
+There are two types of prints, strings and numbers, for that we have printIt and printItNum respectively.
+What these functions are doing is, they use a debug slot defined as 1024, on `debug_slot + 32` we store a value that indicates if the print is going to be a string or a number, then on `debug_slot + 64` we store the value itself, and on `debug_slot` we store a magic value.
+
+Here is were the PrintTracer does its magic, before every execution it looks if the opcode executed is a `HeapWrite`, this is the opcode responsible for storing thing in the Heap, which is in the end what we are doing with the `mstore`, if the value being written is the magic value and its being written on the debug slot, then we know we are in one of the print functions and we need to print the value.
```
Oppen, line 92:
```suggestion
Here is where the PrintTracer does its magic, before every execution it looks if the opcode executed is a `HeapWrite`, this is the opcode responsible for storing things in the Heap, which is in the end what we are doing with the `mstore`, if the value being written is the magic value and its being written on the debug slot, then we know we are in one of the print functions and we need to print the value.
```

juan518munoz (changes requested):
The print tracer example looks great! But I would complete the example by telling which parts of our codebase (besides solidity) we need to change to make it work.

Oppen (approved):

juan518munoz (approved):
Excellent!

PR #101, opened 2024-07-25
Fix array tests

gianbelinche:
This PR fixes tests under `tests/solidity/simple/array`
It also fixes `tests/solidity/simple/algorithm` 

There was an error on FatPointerRead, where if the pointer offset was equal or higher than the height, nothing was done, and the value of the destination was not changed, it is now changed to 0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,4 +22,4 @@ deps: submodules
 	fi
 
 test: deps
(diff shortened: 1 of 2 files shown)


Oppen (approved):

juan518munoz (approved):
LGTM

PR #103, opened 2024-07-25
Optimize CI testing

juan518munoz:
- Current CI runs `make test` which itself does `make deps` this builds a llvm that the CI won't make use of.
- Removing the `deps` dependency from `test` is not possible (makes local tests fail)

Solution is splitting them into two separate `make` commands:
```Makefile
deps: submodules
	cargo install compiler-llvm-builder
	if [ ! -d $(LLVM_PATH) ]; then \
	    cd era-compiler-tester && \
		zksync-llvm clone && zksync-llvm build; \
	fi

# CI test uses LLVM from the era-compiler-llvm repository, doesn't need to build it
ci-test:
	export LLVM_SYS_170_PREFIX=$(LLVM_PATH) && cd era-compiler-tester && cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path  tests/solidity/simple/yul_instructions/ --target EraVM --mode='Y+M3B3 0.8.26'

# Local test uses LLVM from era-compiler-tester submodule, needs to build it
test: deps ci-test
```

This PR also adds `era-compiler-tester` submodule path to `.gitignore`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,12 +14,14 @@ submodules:
 
 deps: submodules
 	cargo install compiler-llvm-builder
-	cd era-compiler-tester && \
-	if [ ! -d "llvm" ]; then \
+	if [ ! -d $(LLVM_PATH) ]; then \
+	    cd era-compiler-tester && \
 		zksync-llvm clone && zksync-llvm build; \
-	else \
-		zksync-llvm build; \
 	fi
 
-test: deps
+# CI test uses LLVM from the era-compiler-llvm repository, doesn't need to build it
+ci-test:
 	export LLVM_SYS_170_PREFIX=$(LLVM_PATH) && cd era-compiler-tester && cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path  tests/solidity/simple/yul_instructions/ --target EraVM --mode='Y+M3B3 0.8.26'
+
+# Local test uses LLVM from era-compiler-tester submodule, needs to build it
+test: deps ci-test
--- a/Makefile
+++ b/Makefile
@@ -19,9 +19,10 @@ deps: submodules
 		zksync-llvm clone && zksync-llvm build; \
 	fi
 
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -14,12 +14,14 @@ submodules:
...
-		zksync-llvm build; \
 	fi
 
-test: deps
+# CI test uses LLVM from the era-compiler-llvm repository, doesn't need to build it
+ci-test:
 	export LLVM_SYS_170_PREFIX=$(LLVM_PATH) && cd era-compiler-tester && cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path  tests/solidity/simple/yul_instructions/ --target EraVM --mode='Y+M3B3 0.8.26'
+
+# Local test uses LLVM from era-compiler-tester submodule, needs to build it
+test: deps ci-test
```
IAvecilla, line 27: Instead of making the test target depending on the ci-test one, we can separate them keeping the same behaviour. I think on something like this:
```
test: deps
     cd era-compiler-tester && cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path tests/solidity/simple/yul_instructions/ --target EraVM --mode='Y+M3B3 0.8.26'
 ```
and for the `ci-test`:
```
ci-test:
   export LLVM_SYS_170_PREFIX=$(LLVM_PATH) && $(MAKE) test
```

juan518munoz: I've run this locally and it worked, I believe that `ci-test` will still run `deps`, but considering that the condition to build llvm is `! -d $(LLVM_PATH)` it wont do so. It will however execute `cargo install compiler-llvm-builder`, but it doesn't consume that much time to be relevant.

gianbelinche: You could add `cargo install compiler-llvm-builder` inside the if then.

gianbelinche (approved):
LGTM

jorbush (approved):
LGTM

PR #97, opened 2024-07-24
Periodical testing

juan518munoz:
Add periodical testing with a cron CI job

PR #105, opened 2024-07-26
Fix revert not returning

gianbelinche:
This PR fixes the following tests:
`tests/solidity/simple/conditional`
`tests/solidity/simple/error`
`tests/solidity/simple/modular`
`tests/solidity/simple/overflow`

Changes output to the following tests:
`tests/solidity/simple/operator`
`tests/solidity/simple/solidity_by_example`
`tests/solidity/simple/yul_semantics`

The revert opcode was not returning any values.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,4 +22,4 @@ deps: submodules
 	fi
 
 test: deps
-	export LLVM_SYS_170_PREFIX=$(LLVM_PATH) && cd era-compiler-tester && cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path  tests/solidity/simple/ --target EraVM
+	export LLVM_SYS_170_PREFIX=$(LLVM_PATH) && cd era-compiler-tester && cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path  tests/solidity/simple/ --target EraVM --mode "Y+M3B3 0.8.26"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -219,7 +219,8 @@ pub fn run(
                     RetOpcode::Revert => match revert(&mut vm, &opcode) {
                         Ok(should_break) => {
                             if should_break {
-                                return Ok((ExecutionOutput::Revert(vec![]), vm));
+                                let result = retrieve_result(&mut vm)?;
+                                return Ok((ExecutionOutput::Revert(result), vm));
                             }
                             Ok(())
                         }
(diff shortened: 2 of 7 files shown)


src/lib.rs
```diff
@@ -251,6 +252,19 @@ pub fn run(
         }
         vm.current_frame_mut()?.pc = opcode_pc_set(&opcode, vm.current_frame()?.pc);
     }
+    let result = retrieve_result(&mut vm)?;
+    Ok((ExecutionOutput::Ok(result), vm))
+}
+
+// Set the next PC according to th enext opcode
```
jorbush, line 259: There is a grammatical error in the comment

jorbush (changes requested):

Makefile
```diff
@@ -22,4 +22,4 @@ deps: submodules
 	fi
 
 test: deps
-	export LLVM_SYS_170_PREFIX=$(LLVM_PATH) && cd era-compiler-tester && cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path  tests/solidity/simple/ --target EraVM
+	export LLVM_SYS_170_PREFIX=$(LLVM_PATH) && cd era-compiler-tester && cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path  tests/solidity/simple/ --target EraVM --mode "Y+M3B3 0.8.26"
```
jorbush, line 25: Why is this flag being added?

src/op_handlers/revert.rs
```diff
@@ -38,6 +38,12 @@ pub fn revert(vm: &mut VMState, opcode: &Opcode) -> Result<bool, EraVmError> {
         }
         Ok(false)
     } else {
+        let register = vm.get_register(opcode.src0_index);
+        let result = get_forward_memory_pointer(register.value, vm, register.is_pointer)?;
+        vm.set_register(
+            opcode.src0_index,
+            TaggedValue::new_pointer(FatPointer::encode(&result)),
+        );
```
jorbush, line 46: We use the same logic in three places, maybe it's time to define a specific function for this. If you don't think it's necessary, I'm fine with that

jorbush (changes requested):

Makefile
gianbelinche: Because without the flag it takes 6 hours to run the CI, which is a lot for PRs.

juan518munoz: We should only run all modes on the [periodical CI job](https://github.com/lambdaclass/era_vm/pull/97) to prevent long CIs for pull requests, it also saves time while testing locally.

jorbush (approved):
LGTM

src/op_handlers/revert.rs
juan518munoz: Adding to this, it may be useful to change:
`if vm.running_contexts.len() > 1 || !vm.current_context()?.near_call_frames.is_empty()`
to:
`if vm.in_far_call() || vm.in_near_call()`
for better clarity across all places where this checks may appear.

Not related to this PR, but might as well add it here

juan518munoz (changes requested):
LGTM will approve once [this comment](https://github.com/lambdaclass/era_vm/pull/105#discussion_r1693244356) is either resolved or refused.

juan518munoz (approved):

PR #107, opened 2024-07-26
Docs: Far Calls vs Near Calls. CallFrames and Context

juan518munoz:


docs/general.md
```diff
@@ -70,9 +72,103 @@ There is no way of modifying heaps via Fat Pointers, they can only be used to re
...
+- Code `Address`
+- Code Page
+- `Stack`
+- Running `CallFrame`s (created by Near Calls)
+- `Heap`
+- `AuxHeap`
+- `CalldataHeap`
+
+The amount of gas that can be allocated to a new `Context` is limited to 63/64 of the currently available gas in the running `Context`.
+
```
gianbelinche, line 92: It would be of a `Callframe`, not a `Context`

docs/general.md
```diff
@@ -70,9 +72,103 @@ There is no way of modifying heaps via Fat Pointers, they can only be used to re
...
+**A new Near Call will inherit the properties of the current `CallFrame`, and make use of the `Stack` and `Heap`s of the running `Context`**.
+
+#### `CallFrame`s are composed of:
+
+- Available gas
+- Exemption handler
```
gianbelinche, line 98: Exception handler

docs/general.md
```diff
@@ -70,9 +72,103 @@ There is no way of modifying heaps via Fat Pointers, they can only be used to re
...
+        libraryContract = LibraryContract(_libraryContractAddress);
+    }
+
+    function callNonReturningFunction() public {
+        libraryContract.someFunction();
+    }
+}
+```
+
+Here we are calling a function of an external contract, this should net us a `far_call` instruction on our compiled code.
```
gianbelinche, line 127: Did you mean set us?

docs/general.md
```diff
@@ -70,9 +72,103 @@ There is no way of modifying heaps via Fat Pointers, they can only be used to re
...
+
+Compiling this code with `zksolc` gives us (part of) the following assembly:
```
gianbelinche, line 129: I think this could be interpreted as:
The code generated by zksolc is a part of this complete code
Its actually the other way around.
Maybe:
> This is part of the assembly compiled with zksolc

docs/general.md
```diff
@@ -70,9 +72,103 @@ There is no way of modifying heaps via Fat Pointers, they can only be used to re
...
+.BB2_2:
+.tmp2:
+	add	 r0, r0, r2
+	ret
+.func_end2:
+...
+```
 
-How far calls pass calldata and returndata between contracts through pointers. Explain the `get_memory_forward_pointer` function and its variants.
+Notice how instead of calling `far_call` directly, we are calling `near_call` which in turn calls `far_call`. This is because `far_call` does not return a value, so we need to wrap it in a `near_call` to return a boolean indicating success wheter the call was successful (1) or not (0).
```
gianbelinche, line 160: This is not completely correct, far calls do have return values, and you store it as a FatPointer to a heap. Its just that with only that value you cannot discern if the call was a success or not. So you need to do this.

gianbelinche (approved):

jorbush (approved):

PR #108, opened 2024-07-29
Initial rollback implementation

gianbelinche:
This PR adds an initial implementation for rollbacks, fixing tests under `solidity/simple/solidity_by_example`

A well thought implementation for rollbacks is needed

### DIFF
--- a/src/op_handlers/revert.rs
+++ b/src/op_handlers/revert.rs
@@ -1,18 +1,16 @@
 use u256::U256;
 
 use crate::{
-    eravm_error::EraVmError,
-    state::VMState,
-    value::{FatPointer, TaggedValue},
-    Opcode,
+    eravm_error::EraVmError, state::VMState, store::Storage, value::{FatPointer, TaggedValue}, Opcode
 };
 
 use super::far_call::{get_forward_memory_pointer, perform_return};
 
-pub fn revert(vm: &mut VMState, opcode: &Opcode) -> Result<bool, EraVmError> {
+pub fn revert(vm: &mut VMState, opcode: &Opcode, storage: &mut dyn Storage) -> Result<bool, EraVmError> {
     vm.flag_eq = false;
     vm.flag_lt_of = false;
     vm.flag_gt = false;
+    storage.rollback(&vm.current_frame()?.storage_before.clone());
     if vm.in_near_call()? {
         let previous_frame = vm.pop_frame()?;
         if opcode.alters_vm_flags {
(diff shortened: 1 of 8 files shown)


src/store.rs
```diff
@@ -15,6 +15,21 @@ pub trait Storage: Debug {
...
+            let value = previous.storage_read(key).unwrap().unwrap();
+            self.storage_write(key, value).unwrap();
+        }
+        let current_keys = self.get_all_keys();
+        for key in current_keys {
+            if previous.storage_read(key).is_err() {
+                self.storage_write(key, U256::zero()).unwrap();
+            }
+        }
+    }
```
Oppen, line 32: I believe it makes more sense to rollback by default, by just dropping the frame, and have a consuming method to apply to the previous state instead. I would also suggest using an `IntoIterator` implementation rather than creating a (potentially huge) vector of keys. The proposed interfaces would be:
```rust
fn into_iter(self) -> impl Iterator<Item = (StorageKey, U256)>; // Or rather, require this as a trait bound
fn apply(self, target: &dyn mut Storage) { // Only called on successful return
    for (k, v) in self.into_iter() { // Can omit the `.into_iter()` if `Self` implements `IntoIterator`
        target.storage_write(k, v);
    }
}
```

gianbelinche: That is a nice approach, I left this PR on draft since I'm currently doing other work, the idea is that doing rollbacks fixes this tests, but It is necessary to think of a better way of performing rollbacks, this could be an option.
We still have the issue of cloning the Storage when implemented with rocksdb, that is another key part to think.

Oppen: We may also avoid the need to `.fake_clone()` simply by making `storage_read()` implementations recursively ask the previous links in the chain. The `struct`s may simply store an optional reference to the previous frame's storage. The number of queries will be bound to the depth of the call stack.

src/store.rs
```diff
@@ -196,6 +217,23 @@ impl Storage for RocksDB {
...
+            let address = H160::from_slice(&key[..20]);
+            let value = U256::from_big_endian(&key[20..]);
+            keys.push(StorageKey::new(address, value));
+            iter.next();
+        }
+        keys
+    }
+    fn fake_clone(&self) -> InMemory {
+        unimplemented!()
+    }
```
Oppen, line 236: For the `RocksDB` logic we may just want to open a [transaction](https://github.com/facebook/rocksdb/wiki/Transactions).

src/op_handlers/revert.rs
```diff
@@ -63,7 +61,8 @@ fn revert_far_call(vm: &mut VMState) -> Result<(), EraVmError> {
     Ok(())
 }
 
-pub fn revert_out_of_gas(vm: &mut VMState) -> Result<(), EraVmError> {
+pub fn revert_out_of_gas(vm: &mut VMState, storage: &mut dyn Storage) -> Result<(), EraVmError> {
+    storage.rollback(&vm.current_frame()?.storage_before.clone());
```
Oppen, line 65: Rolling back on a clone doesn't make a lot of sense to me. Why not just pass by reference?

gianbelinche: Yes that would be fine

MarcosNicolau (approved):

jrchatruc (approved):

PR #128, opened 2024-07-30
Fix `fat_ptr` tests

jorbush:
Fixes the `fat_ptr` tests. 
When it is a pointer, instead of throwing an error, it does nothing, matching the behavior of `vm2`.
Also, I have run all other test suites of  `tests/solidity/simple` and it does not seem to break anything.
Run this command to test these tests:
```sh
cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path tests/solidity/simple/fat_ptr
```


gianbelinche (approved):
LGTM

MarcosNicolau (approved):

JulianVentura (approved):

PR #129, opened 2024-07-30
Fix handling of ERGS costs when expanding heap memory

### LINKED ISSUE #110: Handle ergs cost on `heap_read` on `heap_read.rs`
Ergs cost is being calculated and subtracted from available gas on `heap_read` function, but no revert is being made. Make use of `decrease_gas` function as in `run` function on `lib.rs` with a revert.

MarcosNicolau:
Fixes #110. We were previously mishandling ERGS costs. Now, if the VM runs out of gas during gas cost calculation, we return an error to ensure the error handler triggers a revert.

### DIFF
--- a/src/eravm_error.rs
+++ b/src/eravm_error.rs
@@ -67,4 +67,6 @@ pub enum HeapError {
     StoreOutOfBounds,
     #[error("Trying to read outside of heap bounds")]
     ReadOutOfBounds,
+    #[error("Trying to read/write to heap caused growth which could not be affordable")]
+    NotEnoughErgsToGrowHeap,
 }
--- a/src/op_handlers/aux_heap_read.rs
+++ b/src/op_handlers/aux_heap_read.rs
@@ -22,7 +22,11 @@ pub fn aux_heap_read(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmError
         .get_mut(vm.current_frame()?.aux_heap_id)
         .ok_or(HeapError::ReadOutOfBounds)?
         .expand_memory(addr + 32);
-    vm.current_frame_mut()?.gas_left -= gas_cost;
+
+    let out_of_gas = vm.decrease_gas(gas_cost)?;
+    if out_of_gas {
+        return Err(EraVmError::HeapError(HeapError::NotEnoughErgsToGrowHeap));
+    }
 
     let value = vm
         .heaps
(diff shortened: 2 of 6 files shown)


src/eravm_error.rs
```diff
@@ -67,4 +67,6 @@ pub enum HeapError {
     StoreOutOfBounds,
     #[error("Trying to read outside of heap bounds")]
     ReadOutOfBounds,
+    #[error("Trying to read/write to heap caused growth which could not be affordable")]
```
gianbelinche, line 70: That message seems odd, maybe something like:
Not enough gas to read/write to heap

gianbelinche (changes requested):
Wouldn't it be better for `decrease_gas` to return an error if there is not enough gas instead of returning a boolean?

MarcosNicolau:
> Wouldn't it be better for `decrease_gas` to return an error if there is not enough gas instead of returning a boolean?

Yes, I thought about that too, but given that other modules were using that function, I preferred not to change the current implementation which is already working well. If you think it is worth refactoring, I'll do it!

gianbelinche:
> > Wouldn't it be better for `decrease_gas` to return an error if there is not enough gas instead of returning a boolean?
> 
> Yes, I thought about that too, but given that other modules were using that function, I preferred not to change the current implementation which is already working well. If you think it is worth refactoring, I'll do it!

I think it is, the only place where there could be problems is in `revert_out_of_gas`, the only difference is that in the revert near call, it does not change the heap and aux heap page id, but we already have another issue removing those.

src/lib.rs
```diff
@@ -143,8 +143,7 @@ pub fn run(
             tracer.before_execution(&opcode, &mut vm)?;
         }
 
-        let out_of_gas = vm.decrease_gas(opcode.gas_cost)?;
-        if out_of_gas {
+        if let Some(EraVmError::OutOfGas) = vm.decrease_gas(opcode.gas_cost).err() {
```
gianbelinche, line 146: If another error happens there it would break, match the error, if it is `EraVmError::OutOfGas` do the `revert_out_of_gas` and if it is another do `handle_error`.

gianbelinche (changes requested):

gianbelinche (approved):

jorbush (approved):
LGTM

PR #133, opened 2024-07-30
ci: optimize test pipelines

Oppen:
This PR makes a handful of changes:
1. It splits all the `era-compiler-tester` tests for PRs into directories/files/file+mode sets to allow for parallelization.
2. Uses system rocksdb to speed up compilation.
3. Makes shallow clones of submodules to speed up checkout.
4. Classifies and sorts passing and failing tests. All failing tests appear first.

Future work:
1. Document the process to update the test suite. It requires checking for new and deleted files in the `era-compiler-tests/tests` tree.
2. Document the process to allow passing tests. It should suffice to uncomment the respective line.
3. Try to compile the tester only once and reuse the binary to avoid wasting compute time.
4. Group better the tests, some are rather quick and could be executed together, while some others could take less time.
5. Try to improve caching. `rust-cache` doesn't seem to persist the dependencies, downloading each time, and maybe the system contracts could be cached as well.
6. Check if using the optimization level for the middle-end (M*) as second variable for everything makes jobs fast enough to reduce how many partitions we make. Maybe running one `solidity/simple` job per optimization level gives us good enough times.

Overall, this approach allows running the whole suite in under 17 minutes, which should be good enough for development.

Other notes:
`compiler-tester` provides options to save and load the system contacts. This can be used to cache them, saving about 20s per job. Something similar could be done to the several compilers the tool downloads.


gianbelinche (approved):
LGTM

PR #136, opened 2024-08-01
fix: llvm/stack tests

Oppen:
The primer and the ISA doc state for stack access the offset is computed
by taking the lowest 16 bits.
This commit creates a little trait to easily extract those from `U256`
and `u64`, as well as lowest 32 bits (for stack pointer, that is at most
u16::MAX+1 due to being the biggest stack address + 1).
Doing this fixes a panic when storing to the stack using big values.

Fixes some of the `llvm/stack` tests. The other ones are either buggy or
we need to remove a check that vm2 doesn't (or, possibly, both).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,7 +55,6 @@ jobs:
       fail-fast: false
       matrix:
         testgroup: [
-          #tests/llvm/stack, # FAILING
           #tests/solidity/complex/defi, # FAILING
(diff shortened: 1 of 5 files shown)


gianbelinche (approved):
LGTM

gianbelinche:
Have in mind one check was not successful, it is expected since those tests are being fixed on another PR https://github.com/lambdaclass/era_vm/pull/148#event-13765106378

MarcosNicolau (approved):

PR #144, opened 2024-08-02
Refactor pc increment

### LINKED ISSUE #141: Rethink how to move pc
## Description

After every instruction loop, we increment the `pc` with the `opcode_pc_set` function. On `far_calls`, `near_calls`, `jump`, `ret(Panic)` and `ret(Revert)` opcodes the pc should not get incremented. We are working this around by decreasing the pc by one, except for the `far_call`, where we are actually contemplating that behavior:

```rust
fn opcode_pc_set(opcode: &Opcode, current_pc: u64) -> u64 {
    match opcode.variant {
        Variant::FarCall(_) => 0,
        _ => current_pc + 1,
    }
}
```

Also, whenever there is an `inexplicit_panic`, we can't use the `opcode_pc_set` function, since a `far_call` can actually fail and then pc would not be correctly set as the pc(since we are decrementing it to account for the later increment, [see here](https://github.com/lambdaclass/era_vm/blob/tests-solidity-simple/src/op_handlers/panic.rs#L59)). In such cases we are always incrementing the pc:

```rust
match inexplicit_panic(&mut vm, storage) {
    Ok(false) => {
        vm.current_frame_mut()?.pc += 1;
        continue;
    }
    _ => return Ok((ExecutionOutput::Panic, vm)),
}
```
 
This makes the management of the pc to be spread across the whole codebase, adding unrelated logic inside other functions were we need to account for the pc number.

## Possible solution
My proposed solution is to extend the matching in the op_code_pc_set so that we centralized the pc management and we don't have them all around the codebase. For example:

```rust
// Set the next PC according to the next opcode
fn opcode_pc_set(opcode: &Opcode, current_pc: u64) -> u64 {
    match opcode.variant {
        Variant::FarCall(_) => 0,
        Variant::NearCall(_) => 0,
        Variant::Ret(RetOpcode::Panic) => 0,
        Variant::Ret(RetOpcode::Revert) => 0,
        Variant::Jump(_) => 0,
        _ =>  1,
    }
}
```

MarcosNicolau:
Centralizes the pc increment logic in a single function so that it is not managed across the whole codebases. Closes #141.


src/op_handlers/jump.rs
```diff
@@ -8,7 +8,7 @@ use crate::{opcode::Opcode, state::VMState};
 pub fn jump(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmError> {
     let (src0, _) = address_operands_read(vm, opcode)?;
 
-    let next_pc = (src0.value & U256::from(u64::MAX)) - 1; // we subtract 1 because the pc will be incremented after this function
+    let next_pc = src0.value & U256::from(u64::MAX);
     vm.current_frame_mut()?.pc = next_pc.as_u64();
```
Oppen, line 12: These two could simply be:
```suggestion
    let next_pc = src0.value.low_u64();
    vm.current_frame_mut()?.pc = next_pc;
```

src/lib.rs
```diff
@@ -226,26 +223,33 @@ pub fn run(
...
+// Sets the next PC according to the next opcode
+fn set_pc(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmError> {
+    let current_pc = vm.current_frame()?.pc;
+
+    vm.current_frame_mut()?.pc = match opcode.variant {
         Variant::FarCall(_) => 0,
+        Variant::Ret(RetOpcode::Revert) => current_pc,
+        Variant::Ret(RetOpcode::Panic) => current_pc,
+        Variant::NearCall(_) => current_pc,
+        Variant::Jump(_) => current_pc,
```
juan518munoz, line 248: [Multiple match patterns may be joined with the | operator](https://doc.rust-lang.org/reference/expressions/match-expr.html#:~:text=Multiple%20match%20patterns%20may%20be%20joined%20with%20the%20%7C%20operator.%20Each%20pattern%20will%20be%20tested%20in%20left%2Dto%2Dright%20sequence%20until%20a%20successful%20match%20is%20found.)

```rs
        Variant::Ret(RetOpcode::Revert)
        | Variant::Ret(RetOpcode::Panic)
        | Variant::NearCall(_)
        | Variant::Jump(_) => current_pc,
```

juan518munoz (approved):

gianbelinche (approved):
LGTM

src/lib.rs
MarcosNicolau: Should I change it? This changes in #147 though

juan518munoz: It's just another way to do it, In my opinion you don't need to change it unless you want to.

PR #147, opened 2024-08-05
Fix/refactor ret opcode

MarcosNicolau:
This wasn't breaking any test, but when doing reverts and ok, we were not clearing the registers and the context as specified in the spec. Since ret.panic, ret.revert and ret.ok have very similar behavior, I though it was proper to unify them in ret to make sure we ran the same common behaviour in all of them.


MarcosNicolau:
Depends on #144

juan518munoz:
> when doing reverts and ok, we were not clearing the registers and the context as specified in the spec

Be sure to add a link to the specific part of the spec next time, [like this](https://matter-labs.github.io/eravm-spec/spec.html#:~:text=All%20registers%20are%20cleared).

src/op_handlers/ret.rs
```diff
@@ -0,0 +1,79 @@
...
+
+    if is_failure {
+        storage.rollback(&vm.current_frame()?.storage_before);
+    }
+
+    if vm.in_near_call()? {
+        let previous_frame = vm.pop_frame()?;
+        if use_label && opcode.alters_vm_flags {
+            let to_label = opcode.imm0;
+            vm.current_frame_mut()?.pc = to_label as u64;
```
gianbelinche, line 52: The opcode should only be used if this is an actual Ok, Revert or Panic, if this was due to out of gas or a vm error, then the opcode could be anything and we don't want to use its data

src/op_handlers/ret.rs
```diff
@@ -0,0 +1,79 @@
...
+        let previous_frame = vm.pop_frame()?;
+        vm.current_frame_mut()?.gas_left += previous_frame.gas_left;
+        if is_failure {
+            vm.current_frame_mut()?.pc = previous_frame.exception_handler;
+        } else {
+            vm.current_frame_mut()?.pc += 1;
+        }
+        Ok(false)
+    } else {
+        let result = get_result(vm, opcode, return_type)?;
```
gianbelinche, line 75: Same as above

gianbelinche (changes requested):

gianbelinche (approved):

juan518munoz (approved):

PR #148, opened 2024-08-05
Fix all solidity/simple/function remaining tests

MarcosNicolau:
Fixes all remaining tests for solidity/simple/function. The error was due to storing the stack address destination in registers, which led to unexpected jumps in the contracts.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,531 +1,531 @@
 name: CI
 
 on:
-  merge_group:
-  push:
-    branches: [main]
-  pull_request:
+    merge_group:
+    push:
+        branches: [main]
+    pull_request:
 
 concurrency:
-  group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+    group: ${{ github.workflow }}-${{ github.ref }}
+    cancel-in-progress: true
 
 env:
-  CARGO_TERM_COLOR: always
-  RUST_VERSION: 1.78.0
-  BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
+    CARGO_TERM_COLOR: always
+    RUST_VERSION: 1.78.0
+    BRANCH_NAME: ${{ github.head_ref || github.ref_name }}
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
juan518munoz, line 1: Does this file have any changes besides formatting?

src/state.rs
```diff
@@ -458,11 +458,8 @@ impl Stack {
         &mut self,
         index: usize,
         value: TaggedValue,
-        sp: u16,
+        _sp: u16,
```
juan518munoz, line 461: why keep this argument if it is no longer used?

juan518munoz (changes requested):

src/address_operands.rs
```diff
@@ -112,25 +112,21 @@ fn only_reg_write(
...
-            vm.set_register(opcode.dst0_index, res);
-            res
+            dst0 + TaggedValue::new_raw_integer(U256::from(offset))
         }
         OutputOperandPosition::Second => {
             let dst1 = vm.get_register(opcode.dst1_index);
             let offset = opcode.imm1;
-            let res = dst1 + TaggedValue::new_raw_integer(U256::from(offset));
-            vm.set_register(opcode.dst1_index, res);
-            res
```
gianbelinche, line 133: This `vm.set_register` is actually needed, since when the operand is the second one it always stores on a register, you can check that when this function is called with input `OutputOperandPosition::Second` it is not stored later, maybe it would be better to return the value and store it where it is called

gianbelinche (changes requested):

src/address_operands.rs
```diff
@@ -112,25 +112,21 @@ fn only_reg_write(
...
+            dst1 + TaggedValue::new_raw_integer(U256::from(offset))
```
gianbelinche, line 129: We are never calling this function with `OutputOperandPosition::Second`, we could just remove this argument

JulianVentura (approved):

juan518munoz (approved):

PR #150, opened 2024-08-05
Kernel and Static mode

gianbelinche:
This PR implements kernel and static mode, solving tests under `solidity/complex/interface_casting` and `solidity/complex/nested_calls`


src/state.rs
```diff
@@ -328,7 +331,19 @@ impl VMState {
             Predicate::Le => self.flag_eq || self.flag_lt_of,
             Predicate::Ne => !self.flag_eq,
             Predicate::GtOrLt => self.flag_gt || self.flag_lt_of,
+        };
+        let mut kernel_ok = true;
```
MarcosNicolau, line 335: I would do `kernel_ok =  if opcode.variant.requires_kernel_mode() { self.current_context()?.is_kernel() } else { true }`. But it is just a biased opinion, you don't have to change it. Same applies for the context

src/op_handlers/ret.rs
```diff
@@ -25,6 +25,9 @@ fn get_result(
     }
     let register = vm.get_register(reg_index);
     let result = get_forward_memory_pointer(register.value, vm, register.is_pointer)?;
+    if !vm.current_context()?.is_kernel() && result.page == vm.current_context()?.calldata_heap_id {
+        return Err(EraVmError::NonValidForwardedMemory);
```
MarcosNicolau, line 29: Are you sure you have to return an err here? I think in this case you should just return a new `TaggedValue::new_pointer(U256::zero())`

src/op_handlers/far_call.rs
```diff
@@ -204,7 +204,9 @@ pub fn far_call(
 
     let exception_handler = opcode.imm0 as u64;
 
-    let abi = get_far_call_arguments(src0.value);
+    let mut abi = get_far_call_arguments(src0.value);
+    abi.is_constructor_call = abi.is_constructor_call && vm.current_context()?.is_kernel();
+    abi.is_system_call = abi.is_system_call && is_kernel(&contract_address);
```
juan518munoz, line 209: Is there any reason why lines `208` and `209` are not being handled inside of `get_far_call_arguments()`?

src/op_handlers/ret.rs
gianbelinche: On the vm2 they do a panic 
```
let result = get_far_call_calldata(raw_abi, is_pointer, vm, false).filter(|pointer| {
                vm.state.current_frame.is_kernel
                    || pointer.memory_page != vm.state.current_frame.calldata_heap
            });

            if result.is_none() {
                return_type = ReturnType::Panic;
            }
```
Doing this will call the `inexplicit_panic` that performs the correct behaviour

src/op_handlers/far_call.rs
gianbelinche: Not really, just that the vm2 does it this way, however that would involve passing the vm and the contract address as parameters of `get_far_call_arguments`, i think is simpler this way

juan518munoz (approved):
LGTM

PR #152, opened 2024-08-05
Fix solidity/complex/yul instructions/call

MarcosNicolau:
Fixes all tests under solidity/complex/yul_instructions/call for all modes

PR #153, opened 2024-08-06
Fix complex/yul instructions staticcall

MarcosNicolau:
They were failing because we were not correctly panicking when checking static mode.


MarcosNicolau:
depends on #152

jorbush (approved):
LGTM

gianbelinche (approved):
LGTM

src/lib.rs
```diff
@@ -117,14 +117,15 @@ pub fn run(
...
+        let can_execute = vm.can_execute(&opcode);
+        if vm.decrease_gas(opcode.gas_cost).is_err() || can_execute.is_err() {
             match inexplicit_panic(&mut vm, storage) {
                 Ok(false) => continue,
                 _ => return Ok((ExecutionOutput::Panic, vm)),
             }
         }
 
-        if vm.can_execute(&opcode)? {
+        if can_execute.unwrap() {
```
juan518munoz, line 128: Handle error:
```suggestion
        if can_execute? {
```

juan518munoz (changes requested):

juan518munoz:
Ignore requested changes, I missed the part above where `is_err()` is checked, so you don't need to `?` the `Result`.

You can still apply the suggestion to prevent having `unwrap`s on our codebase, but it's up to you.

src/lib.rs
MarcosNicolau: In theory, it shouldn't return err since we are already checking if is_err above, but I'll change it just to make sure

juan518munoz (approved):

PR #154, opened 2024-08-06
Fix complex/yul instructions/codecopy-calldatacopy-calldataload

MarcosNicolau:
Fixes all remaining tests under:
- `solidity/complex/yul instructions/codecopy`
- `solidity/complex/yul instructions/calldatacopy`
- `solidity/complex/yul instructions/calldataload`

They were failing because: 
- we were not validating if the address could overflow when doing `fat_ptr read`,
- we were not properly validating if the operands sources were larger than u32 when fetching ptr operators.

### DIFF
--- a/src/op_handlers/far_call.rs
+++ b/src/op_handlers/far_call.rs
@@ -102,10 +102,10 @@ fn far_call_params_from_register(
     let mut ergs_passed = source.0[3] as u32;
     let gas_left = vm.gas_left()?;
 
-    if ergs_passed > gas_left {
-        ergs_passed = (gas_left * FAR_CALL_GAS_SCALAR_MODIFIER_DIVIDEND)
-            / FAR_CALL_GAS_SCALAR_MODIFIER_DIVISOR;
-    }
+    let maximum_gas =
+        gas_left / FAR_CALL_GAS_SCALAR_MODIFIER_DIVISOR * FAR_CALL_GAS_SCALAR_MODIFIER_DIVIDEND;
+    ergs_passed = ergs_passed.min(maximum_gas);
+
     source.to_little_endian(&mut args);
     let [.., shard_id, constructor_call_byte, system_call_byte] = args;
 
--- a/src/op_handlers/fat_pointer_read.rs
+++ b/src/op_handlers/fat_pointer_read.rs
@@ -10,6 +10,13 @@ pub fn fat_pointer_read(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmEr
     if !src0.is_pointer {
         return Err(OperandError::InvalidSrcNotPointer(opcode.variant).into());
     }
+
+    // make sure the pointer is a value that can be added without overflow for 32 bits.
+    if src0.value.low_u32() > (u32::MAX - 32) {
+        let _ = vm.set_gas_left(0);
+        return Err(HeapError::InvalidAddress.into());
+    }
+
     let pointer = FatPointer::decode(src0.value);
 
     let value = if pointer.offset < pointer.len {
--- a/src/op_handlers/fat_pointer_read.rs
+++ b/src/op_handlers/fat_pointer_read.rs
@@ -13,7 +13,7 @@ pub fn fat_pointer_read(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmEr
 
     // make sure the pointer is a value that can be added without overflow for 32 bits.
     if src0.value.low_u32() > (u32::MAX - 32) {
-        let _ = vm.set_gas_left(0);
+        vm.set_gas_left(0)?;
         return Err(HeapError::InvalidAddress.into());
     }
 
--- a/src/ptr_operator.rs
+++ b/src/ptr_operator.rs
@@ -1,5 +1,3 @@
-use zkevm_opcode_defs::MAX_OFFSET_FOR_ADD_SUB;
-
 use crate::{
     address_operands::{address_operands_read, address_operands_store},
     eravm_error::{EraVmError, OperandError},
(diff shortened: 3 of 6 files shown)


MarcosNicolau:
depends on #153

MarcosNicolau:
This single test is failing in various modes: FAILED tests/solidity/complex/yul_instructions/codecopy/test.json::ordinar_words_size[test:1]

I suspect the issue is actually the same for all the failed jobs.

src/ptr_operator.rs
```diff
@@ -19,7 +17,7 @@ pub fn ptr_operands_read(
     }
 
     let pointer = FatPointer::decode(src0.value);
-    if src1.value > MAX_OFFSET_FOR_ADD_SUB {
+    if src1.value > u32::MAX.into() {
```
juan518munoz, line 20: `MAX_OFFSET_FOR_ADD_SUB == u32::MAX` 

Am I missing something?

src/op_handlers/fat_pointer_read.rs
```diff
@@ -10,6 +10,13 @@ pub fn fat_pointer_read(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmEr
     if !src0.is_pointer {
         return Err(OperandError::InvalidSrcNotPointer(opcode.variant).into());
     }
+
+    // make sure the pointer is a value that can be added without overflow for 32 bits.
+    if src0.value.low_u32() > (u32::MAX - 32) {
+        let _ = vm.set_gas_left(0);
```
juan518munoz, line 16: Handle error:
```rs
vm.set_gas_left(0)?;
```

juan518munoz (changes requested):
Looks good 🚀 , but please address review comments.

src/ptr_operator.rs
MarcosNicolau: Yes I know, but for some reason that was not working

src/op_handlers/fat_pointer_read.rs
MarcosNicolau: In their implementation, they burn all gas as well. See [here](https://github.com/matter-labs/vm2/blob/master/src/instruction_handlers/heap_access.rs#L68).

MarcosNicolau: We don't return the err, because we will return an err anyways, and we want to specify that it is because the address was invalid

src/ptr_operator.rs
gianbelinche: `MAX_OFFSET_FOR_ADD_SUB` is `u32:MAX + 1`

juan518munoz: @gianbelinche is correct

src/op_handlers/fat_pointer_read.rs
juan518munoz: Burning all gas seems correct.

Their [`use_gas(&mut self, amount: u32)`](https://github.com/matter-labs/vm2/blob/9a38900d7af9b1d72b47ce3be980e77c1239a61d/src/state.rs#L82-L83) function will only return an `Err` if `self.current_frame.gas < amount`, and even then it will set the current gas to zero, that's why they don't need to handle the error.

Our gas setting function may also fail due to a `Vec<>` access error (see `VMState::current_context_mut`).

MarcosNicolau: You are right!

juan518munoz (approved):

src/op_handlers/far_call.rs
```diff
@@ -102,10 +102,10 @@ fn far_call_params_from_register(
     let mut ergs_passed = source.0[3] as u32;
     let gas_left = vm.gas_left()?;
 
-    if ergs_passed > gas_left {
-        ergs_passed = (gas_left * FAR_CALL_GAS_SCALAR_MODIFIER_DIVIDEND)
-            / FAR_CALL_GAS_SCALAR_MODIFIER_DIVISOR;
-    }
+    let maximum_gas =
+        gas_left / FAR_CALL_GAS_SCALAR_MODIFIER_DIVISOR * FAR_CALL_GAS_SCALAR_MODIFIER_DIVIDEND;
```
Oppen, line 106: Careful here. If `gas_left < 64` this will be `0`, while it should be in almost all cases non-`0`. For example, `63 * 63/64 == 62.015625`, which truncates to `62`, would result in `0` if you perform the integer division first.

src/op_handlers/fat_pointer_read.rs
```diff
@@ -10,6 +10,13 @@ pub fn fat_pointer_read(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmEr
     if !src0.is_pointer {
         return Err(OperandError::InvalidSrcNotPointer(opcode.variant).into());
     }
+
+    // make sure the pointer is a value that can be added without overflow for 32 bits.
+    if src0.value.low_u32() > (u32::MAX - 32) {
+        vm.set_gas_left(0)?;
+        return Err(HeapError::InvalidAddress.into());
```
Oppen, line 17: Maybe we should `panic_from_far_call` here? Does that helper take the error code to forward it?

MarcosNicolau: Yes, any err returned from far_call will trigger the`panic_from_far_call`. See [here](https://github.com/lambdaclass/era_vm/blob/main/src/lib.rs#L179C17-L185C24).

src/op_handlers/far_call.rs
MarcosNicolau: I see that could be an issue. The thing is, both implementations (vm1, vm2) do it this way.

Oppen (approved):

src/op_handlers/far_call.rs
MarcosNicolau: [See](https://github.com/matter-labs/vm2/blob/master/src/instruction_handlers/far_call.rs#L97-#L100)

PR #160, opened 2024-08-06
Evm interpreter

gianbelinche:


src/state.rs
```diff
@@ -550,6 +551,16 @@ impl Heap {
         result
     }
 
+    pub fn read_unaligned_from_pointer(&self, pointer: &FatPointer) -> Vec<u8> {
+        let mut result = Vec::new();
+        let start = pointer.start + pointer.offset;
+        let finish = start + pointer.len;
+        for i in start..finish {
+            result.push(self.heap[i as usize]);
```
fkrause98, line 559: This is assuming the heap is always sized properly, right? This will panic if we're out of bounds, let's check that the bounds are correct or either return an error ir self.heap.get(i) is None.

src/tracers/blob_saver_tracer.rs
```diff
@@ -0,0 +1,102 @@
...
+    0x00, 0x00, 0x80, 0x06,
+]);
+
+const KNOWN_CODES_STORAGE_ADDRESS: Address = H160([
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x80, 0x04,
+]);
+
+pub(crate) fn hash_evm_bytecode(bytecode: &[u8]) -> H256 {
+    use zkevm_opcode_defs::sha2::{Digest, Sha256};
```
fkrause98, line 31: Please put imports at the top of the file.

src/tracers/blob_saver_tracer.rs
```diff
@@ -0,0 +1,102 @@
...
+    output[..].copy_from_slice(result.as_slice());
+    output[0] = 2; //BLOB version byte
+    output[1] = 0;
+    output[2..4].copy_from_slice(&len.to_be_bytes());
+
+    H256(output)
+}
+
+fn bytes_to_be_words(vec: &[u8]) -> Vec<U256> {
+    assert!(vec.len() % 32 == 0, "Invalid bytecode length");
```
fkrause98, line 47: This will panic if `vec.len() % 32 != 0`, let's make it return an error if that happens.

src/tracers/blob_saver_tracer.rs
```diff
@@ -0,0 +1,102 @@
...
+        if data.len() < 4 {
+            // Not interested
+            return Ok(());
+        }
+
+        let (signature, data) = data.split_at(4);
+
+        // Hardcoded signature of publishEVMBytecode function
+        // in hex that is 0x964eb607
+        if signature != [150, 78, 182, 7] {
```
fkrause98: Nit-pick: Given the comment, better to use `[0x96, 0x4e, 0xb6, 0x7]` here instead then, for consistency.

src/vm.rs
```diff
@@ -67,18 +69,26 @@ impl EraVM {
...
-    fn run_opcodes(&mut self) -> ExecutionOutput {
-        self.run(&mut [], EncodingMode::Production)
-            .unwrap_or(ExecutionOutput::Panic)
+    fn run_opcodes(&mut self) -> (ExecutionOutput, BlobSaverTracer) {
+        let mut tracer = BlobSaverTracer::new();
+        let r = self
+            .run(&mut [Box::new(&mut tracer)], EncodingMode::Production)
+            .unwrap_or(ExecutionOutput::Panic);
+
+        (r, tracer)
```
gianbelinche, line 91: Instead of creating the tracer and returning it, the idea is passing any tracer you want from outside and the consuming it if needed, that means,  passing the tracer from compiler tester

src/state.rs
```diff
@@ -550,6 +551,16 @@ impl Heap {
         result
     }
 
+    pub fn read_unaligned_from_pointer(&self, pointer: &FatPointer) -> Vec<u8> {
```
gianbelinche, line 554: Why doesn't this work with the normal read from pointer?

src/tracers/blob_saver_tracer.rs
JulianVentura: I modified the code a little bit since you made this comment. Now there is a check but we are not returning an error if the bytecode len is not multiple of 32. Instead, we return an `Ok(())` and log an error. Do you think it's better return an error from the tracer and stop execution?

src/tracers/blob_saver_tracer.rs
JulianVentura: Sure!

src/op_handlers/opcode_decommit.rs
```diff
@@ -0,0 +1,45 @@
...
+        .decommit(code_hash)?
+        .ok_or(EraVmError::DecommitFailed)?;
+
+    let id = vm.heaps.allocate();
+    let heap = vm.heaps.get_mut(id).ok_or(HeapError::StoreOutOfBounds)?;
+    let mut address = 0;
+    for value in code.iter() {
+        heap.store(address, *value);
+        address += 32;
+    }
```
gianbelinche, line 33: Im the one who did this but...
Maybe we need to do a `heap.expand_memory` here and consume the gas needed

src/state.rs
JulianVentura: Because this returns a `Vec<u8>` instead of a `U256`.

src/tracers/blob_saver_tracer.rs
JulianVentura: Done

src/state.rs
JulianVentura: Done

src/tracers/blob_saver_tracer.rs
```diff
@@ -0,0 +1,112 @@
...
+const KNOWN_CODES_STORAGE_ADDRESS: Address = H160([
+    0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+    0x00, 0x00, 0x80, 0x04,
+]);
+
+// Hardcoded signature of `publishEVMBytecode` function.
+// In hex is 0x964eb607
+const PUBLISH_BYTECODE_SIGNATURE: [u8; 4] = [0x96, 0x4e, 0xb6, 0x7];
+
+pub(crate) fn hash_evm_bytecode(bytecode: &[u8]) -> H256 {
```
MarcosNicolau, line 40: is there any reason we need `pub(crate)` here?

src/state.rs
MarcosNicolau: I believe the intended behaviour is to break the loop instead.

MarcosNicolau (changes requested):

src/state.rs
gianbelinche: You can create the `Vec<u8>` from the `U256` (or viceversa), I don't think it is a good a idea to have to functions that do almost the same, in any case, replace the implementation of one of them with the other plus a conversion

src/op_handlers/opcode_decommit.rs
JulianVentura: Done

MarcosNicolau (approved):

src/tracers/blob_saver_tracer.rs
JulianVentura: Done

JulianVentura (approved):

PR #161, opened 2024-08-06
ci: avoid repeated compilation and downloads

Oppen:
Split the test job in a compilation+download stage and a test stage.
The first stage installs dependencies and builds the tester for this
run, and then packs everything up and uploads it as cache.
The second stage then downloads that bundle and uses that to run the
tests. It's marked as dependent on the first stage, so it should always
find a populated cache.


MarcosNicolau (approved):
Excelent!! 👍

fkrause98 (approved):

PR #163, opened 2024-08-06
Rename VM flags

### LINKED ISSUE #151: Change flag names
Currently when parsing an opcode we do the following:

```
let [alters_vm_flags, swap_flag] = match opcode_zksync.opcode {
       Variant::Ptr(_) => [false, opcode_zksync.flags[0]],
       _ => opcode_zksync.flags,
};
```

Naming `alters_vm_flags` and `swap_flags` respectively, however different opcodes use this flags for different things, for example far call uses alters_vm_flags for checking if the new context should be static, and ret uses it if it uses a label.

We should change this to something like flag 0 and flag 1, and indicate in each opcode what it means

juan518munoz:
Closes #151 

Current naming scheme for the two flags in the vm do not represent their actual use in some cases (see linked issue for more details).

### DIFF
--- a/src/address_operands.rs
+++ b/src/address_operands.rs
@@ -87,7 +87,7 @@ pub fn address_operands_read(
             }
         }
     };
-    if opcode.swap_flag {
+    if opcode.flag1_set {
         Ok((op2, op1))
     } else {
         Ok((op1, op2))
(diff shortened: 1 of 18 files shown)


Oppen (approved):

MarcosNicolau (approved):
LGTM 👍

PR #164, opened 2024-08-06
Fix solidity/ethereum/inlineAssembly

MarcosNicolau:
They were failing because we were not properly storing the values in the transient storage between frames. Also, in the `fat_pointer_read` opcode we were incorrectly reading the values.

### DIFF
--- a/src/op_handlers/ret.rs
+++ b/src/op_handlers/ret.rs
@@ -31,18 +31,19 @@ fn get_result(
     Ok(TaggedValue::new_pointer(FatPointer::encode(&result)))
 }
 
-fn save_transient_store(vm: &mut VMState, prev_storage: InMemory) {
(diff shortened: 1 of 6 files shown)


Oppen (approved):

src/op_handlers/ret.rs
```diff
@@ -31,6 +31,20 @@ fn get_result(
     Ok(TaggedValue::new_pointer(FatPointer::encode(&result)))
 }
 
+fn save_transient_store(vm: &mut VMState, prev_storage: InMemory) {
```
fkrause98, line 34: Let's make this return a `Result<(), EraVMError>`, also rewrite these unwraps to proper error handling or use `?`,.

fkrause98 (changes requested):
Great job! Let's discuss some changes and we're done here.

JulianVentura (approved):
LGTM

fkrause98 (approved):

PR #166, opened 2024-08-07
Implement aux_mutating0 opcode

MarcosNicolau:
This opcode actually does nothing, shouldn't panic though, just increment pc. See [here](https://github.com/matter-labs/era-zk_evm/blob/v1.5.0/src/opcodes/execution/context.rs#L42C1-L47C1) and [here](https://github.com/matter-labs/vm2/blob/master/src/instruction_handlers/context.rs#L115-L123).

### DIFF
--- a/src/op_handlers/context.rs
+++ b/src/op_handlers/context.rs
@@ -77,7 +77,8 @@ pub fn set_context_u128(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmEr
 }
 
 pub fn aux_mutating0(_vm: &mut VMState, _opcode: &Opcode) -> Result<(), EraVmError> {
-    panic!("Not yet implemented");
+    // does nothing for now
+    Ok(())
 }
 
 pub fn increment_tx_number(vm: &mut VMState, _opcode: &Opcode) -> Result<(), EraVmError> {


gianbelinche (commented):
Was this breaking any test?
If not I think it is better to leave it as it was.
It is a lot easier to spot a failed test if it just panics with `not_yet_implemented`
Maybe change it to an `unimplemented! `

MarcosNicolau:
Not really, I was addressing issue #166.

JulianVentura (commented):
I agree with Gian. I think we should leave it like it is since it's easier to spot a bug if it just panics.

src/op_handlers/context.rs
```diff
@@ -77,7 +77,8 @@ pub fn set_context_u128(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmEr
 }
 
 pub fn aux_mutating0(_vm: &mut VMState, _opcode: &Opcode) -> Result<(), EraVmError> {
-    panic!("Not yet implemented");
+    // does nothing for now
+    Ok(())
```
fkrause98, line 81:
```suggestion
unimplemented!("Not implemented on any VMs")
```

I'd leave it like this.

fkrause98 (changes requested):

src/op_handlers/context.rs
fkrause98, line 81: Disregard my last comment, calling this function should return an `Err(...)` much like an invalid opcode on lib.rs does, you can put the error here or on the matching branch for this instruction over on lib.rs

fkrause98 (changes requested):

src/op_handlers/context.rs
MarcosNicolau: That, wouldn't be compatible with their current behaviour. Both of their implementation just ignore the opcode and increment the pc. Returning err would trigger a ret.panic. Currently, none of the contracts we have tested falls here, so maybe a panic would be ok to spot it later, but in theory this opcode should just increment the pc.

fkrause98 (approved):

src/op_handlers/context.rs
gianbelinche: I think the best idea would be to return an `ExecutionOutput::Panic` disregarding any type of frame pop or anything, just end execution, if we just leave it with an ok, and someday someone adds a test that uses it it will be really hard to catch, but if we returned a Panic it would be easier to see that the last opcode executed was this one

MarcosNicolau:
Closing in favor of #172.

PR #168, opened 2024-08-07
Run compiler tester single threaded

Oppen:
This could help avoid OOM situations. We will still need to find where our memory usage grew, but this should enable iteration.


gianbelinche (approved):
LGTM

PR #172, opened 2024-08-07
Refactor unimplemented opcodes

MarcosNicolau:
For those opcodes that are not yet implemented, instead of panicking the process, stop vm execution as panic.


src/op_handlers/unimplemented.rs
```diff
@@ -0,0 +1,7 @@
+use crate::eravm_error::{EraVmError, OpcodeError};
+use crate::{opcode::Opcode, state::VMState};
+
+pub fn unimplemented(_vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmError> {
+    eprintln!("Unimplemented instruction: {:?}!", opcode.variant);
```
jorbush, line 5: why is this `print` necessary?

src/op_handlers/context.rs
```diff
@@ -76,7 +76,7 @@ pub fn set_context_u128(vm: &mut VMState, opcode: &Opcode) -> Result<(), EraVmEr
     Ok(())
 }
 
-pub fn aux_mutating0(_vm: &mut VMState, _opcode: &Opcode) -> Result<(), EraVmError> {
+pub fn _aux_mutating0(_vm: &mut VMState, _opcode: &Opcode) -> Result<(), EraVmError> {
```
jorbush, line 79: Seeing the new implementation of `unimplemented` this no longer seems to be necessary

jorbush (changes requested):

src/op_handlers/unimplemented.rs
MarcosNicolau: Just to make sure we can easier spot where the execution ended

src/op_handlers/context.rs
MarcosNicolau: I left it for when we have to implement it

jorbush (approved):
LGTM

Oppen (approved):

PR #173, opened 2024-08-07
Implement precompiles

### LINKED ISSUE #94: Get rid of the `zk_evm_abstractions` dependency
We are using [era-zk_evm_abstractions](https://github.com/matter-labs/era-zk_evm_abstractions) as a dependency to implement the precompile opcode. Removing it means

- Implementing the `keccak256`, `sha256`, `ecRecover` and `SECP256R1` algorithms ourselves or getting them from another library.
- Calling those algorithms and storing their result in the appropriate Heap when the precompile opcode is called (going through the evm abstractions code is probably the easiest way to understand how this works).

### LINKED ISSUE #121: Implement `Memory` trait fully for `Heaps` or remove it
Currently, we have a partial implementation of `trait Memory` for struct `Heaps`, panicking if `specialized_code_query` and `read_code_query` are called. We should implement those two methods or delete the trait implementation if it is not being used.

juan518munoz:
This PR implements precompiles, with the main purpose of no longer depending in  the `zk_evm_abstractions` crate.

Closes #121 
Closes #94


MarcosNicolau:
`ecrecover` and `secp256r1` implemented!

src/precompiles/ecrecover.rs
```diff
@@ -0,0 +1,230 @@
...
+
+        r_value.to_big_endian(&mut buffer[..]);
+        let r_bytes = buffer;
+
+        s_value.to_big_endian(&mut buffer[..]);
+        let s_bytes = buffer;
+
+        v_value.to_big_endian(&mut buffer[..]);
+        let v = buffer[31];
+        assert!(v == 0 || v == 1);
```
fkrause98, line 74: Can't we return an error here?

fkrause98, line 74: Can't we return an error here?

src/precompiles/ecrecover.rs
```diff
@@ -0,0 +1,230 @@
...
+
+    recover_no_malleability_check(digest, signature, recid)
+}
+
+fn recover_no_malleability_check(
+    digest: &[u8; 32],
+    signature: k256::ecdsa::Signature,
+    recovery_id: k256::ecdsa::RecoveryId,
+) -> Result<VerifyingKey, ()> {
+    use k256::ecdsa::hazmat::bits2field;
```
fkrause98, line 168: Please move these imports to the top of the file.

src/heaps.rs
```diff
@@ -46,4 +50,25 @@ impl Heaps {
     pub fn get_mut(&mut self, index: u32) -> Option<&mut Heap> {
         self.heaps.get_mut(index as usize)
     }
+
+    pub fn execute_partial_query(
```
fkrause98, line 54: I know this can be a bit hard to do but -- can we add some comments here on what we think this function does? It does not matter if it's precise or not, but we should start having an understanding on this and memory queries, feel free to check the original repo if you have to.

src/precompiles/ecrecover.rs
```diff
@@ -0,0 +1,230 @@
...
+        let params = precompile_abi_in_log(precompile_call_params);
+
+        let mut current_read_location = MemoryLocation {
+            page: params.memory_page_to_read,
+            index: params.output_memory_offset,
+        };
+
+        let hash_query = MemoryQuery {
+            location: current_read_location,
+            value: U256::zero(),
```
fkrause98, line 24: Do we know why this MemoryQuery needs a zero?

src/precompiles/ecrecover.rs
```diff
@@ -0,0 +1,230 @@
...
+                value_is_pointer: false,
+                rw_flag: true,
+            };
+            heaps.execute_partial_query(result_query)?;
+        }
+        Ok(())
+    }
+}
+
+pub fn ecrecover_inner(
```
fkrause98, line 145: Same as above, let's try and add a comment -- however small -- on what we think this function does. Seems like this is Elliptic Curve Recover of a pair, but feel free to correct me.

src/precompiles/secp256r1_verify.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+pub fn secp256r1_verify_inner(
+    digest: &[u8; 32],
+    r: &[u8; 32],
+    s: &[u8; 32],
+    x: &[u8; 32],
+    y: &[u8; 32],
+) -> Result<bool, ()> {
+    use p256::ecdsa::signature::hazmat::PrehashVerifier;
```
fkrause98, line 147: Please move this imports to the top of the file.

src/precompiles/sha256.rs
```diff
@@ -0,0 +1,109 @@
...
+                round_witness.writes = Some([result_query]);
+            }
+        }
+        Ok(())
+    }
+}
+
+pub fn sha256_rounds_function(abi_key: U256, heaps: &mut Heaps) -> Result<(), EraVmError> {
+    let mut processor = Sha256Precompile;
+    processor.execute_precompile(abi_key, heaps)
```
fkrause98, line 108: Nit-pick here, can't we just do...?
```suggestion
    Sha256Precompile.execute_precompile(abi_key, heaps)
```

fkrause98 (changes requested):
Looks good, but let's discuss some things.

src/precompiles/ecrecover.rs
MarcosNicolau: It is the default value, it gets replaced when executing the queries.

src/precompiles/sha256.rs
juan518munoz: Good suggestion! I've applied it for `Keccak256Precompile` as well.

src/precompiles/ecrecover.rs
fkrause98: Super.

fkrause98 (approved):
Great job! 🚀

src/precompiles/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+            value_is_pointer: false,
+        }
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
+pub struct MemoryLocation {
+    pub page: u32,
+    pub index: u32,
+}
```
gianbelinche, line 36: Can't we do this without reimplementing their way of accessing heap? I think the idea was to stop using this kind of access, and not just reimplementing it for our heap, if we have the page and offset we could just read from the heap with our current implementations.
But correct me if that is not the case

gianbelinche (changes requested):

Cargo.toml
```diff
@@ -6,9 +6,12 @@ edition = "2021"
...
 hex = "0.4.3"
 u256 = { package = "primitive-types", version = "0.12.1" }
+k256 = "0.13.3"
 zkevm_opcode_defs = { git = "https://github.com/matter-labs/era-zkevm_opcode_defs.git", branch = "v1.5.1" }
-zk_evm_abstractions = { git = "https://github.com/matter-labs/era-zk_evm_abstractions.git", branch = "v1.5.1" }
-# zk_evm_abstractions = { path = "../era-zk_evm_abstractions" }
 rocksdb = "0.21.0"
 thiserror = "1.0.61"
+sha2 = "0.11.0-pre.0"
+sha3 = "0.11.0-pre.0"
```
Oppen, line 14: I'm not sure using a pre-release is a good idea.

src/precompiles/keccak256.rs
```diff
@@ -0,0 +1,194 @@
...
+        meaningful_bytes: usize,
+    ) {
+        assert!(self.filled + meaningful_bytes <= BUFFER_SIZE);
+        self.bytes[self.filled..(self.filled + meaningful_bytes)]
+            .copy_from_slice(&input[offset..(offset + meaningful_bytes)]);
+        self.filled += meaningful_bytes;
+    }
+
+    pub fn consume<const N: usize>(&mut self) -> [u8; N] {
+        assert!(N <= BUFFER_SIZE);
```
Oppen, line 35: Considering this is known at compile time, maybe assert this at compile time as well:
```suggestion
        const { assert!(N <= BUFFER_SIZE) };
```
I **think** this is legal to do and it forces build time evaluation of the assertion.

src/precompiles/keccak256.rs
```diff
@@ -0,0 +1,194 @@
...
+        self.filled + num_bytes <= BUFFER_SIZE
+    }
+
+    pub fn fill_with_bytes<const N: usize>(
+        &mut self,
+        input: &[u8; N],
+        offset: usize,
+        meaningful_bytes: usize,
+    ) {
+        assert!(self.filled + meaningful_bytes <= BUFFER_SIZE);
```
Oppen, line 28: Should we also check that `N <= BUFFER_SIZE`in this function as well?

src/precompiles/keccak256.rs
```diff
@@ -0,0 +1,194 @@
...
+
+fn get_state_bytes(bytes: &[u8]) -> [u64; 25] {
+    let mut result = [0u64; 25];
+
+    // grab in chunks of 4 bytes and reverse them
+    for i in 0..25 {
+        let mut chunk = [0u8; 8];
+        chunk.copy_from_slice(&bytes[(i * 8)..((i + 1) * 8)]);
+        result[i] = u64::from_le_bytes(chunk);
+    }
```
Oppen, line 59: The comment contradicts the code here. It's 8 bytes chunks, not 4 bytes. Which one is correct?

src/precompiles/ecrecover.rs
```diff
@@ -0,0 +1,155 @@
...
+impl Precompile for ECRecoverPrecompile {
+    fn execute_precompile(&mut self, query: U256, heaps: &mut Heaps) -> Result<(), EraVmError> {
+        let precompile_call_params = query;
+        let params = precompile_abi_in_log(precompile_call_params);
+        let read_heap_idx = params.memory_page_to_read;
+        let write_heap_idx = params.memory_page_to_write;
+        let mut addr = params.output_memory_offset;
+
+        let (hash_value, _) =
+            heaps.read(params.memory_page_to_read, params.output_memory_offset)?;
```
Oppen, line 30:
```suggestion
        let (hash_value, _) = heaps.read(read_heap_idx, addr)?;
```

src/precompiles/keccak256.rs
```diff
@@ -0,0 +1,164 @@
...
+        }
+
+        let source_memory_page = params.memory_page_to_read;
+        let destination_memory_page = params.memory_page_to_write;
+        let write_offset = params.output_memory_offset;
+
+        let mut input_buffer = ByteBuffer::<KECCAK_PRECOMPILE_BUFFER_SIZE> {
+            bytes: [0u8; KECCAK_PRECOMPILE_BUFFER_SIZE],
+            filled: 0,
+        };
```
Oppen, line 92: You can just derive `Default` for `ByteBuffer`. It'll save some typing.

Cargo.toml
fkrause98: Good catch, can we change this @juan518munoz ?

MarcosNicolau: We are trying to change them, but we are having some issues when getting the intermediate state of the hasher. We were using the crypto commons crate for that, but the function to do so is only available on the pre release version.

src/state.rs
```diff
@@ -529,12 +529,19 @@ impl Heap {
...
         }
         result
     }
 
+    pub fn expanded_read(&mut self, address: u32) -> (U256, u32) {
+        let gas_cost = self.expand_memory(address + 32);
+        let result = self.read(address);
+        (result, gas_cost)
+    }
+
```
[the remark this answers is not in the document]
MarcosNicolau: Yes I was thinking of it as well, but I didn't want to touch unrelated stuff in this pr. When this one is merged, I'll open a new one!

src/precompiles/mod.rs
```diff
@@ -0,0 +1,80 @@
...
+
+fn get_hasher_state<const N: usize, T, H>(hasher: H) -> T {
+    // casts the hasher ptr to the CoreWrapper struct
+    let raw_ptr = &hasher as *const _ as *const CoreWrapper<T, N>;
+    // this is not unsafe since we are replicating the structure(thus same layout) of the original ptr
+    // this a hack that allows us to access private fields
+    // also ptr::read does not moved the value from memory
+    unsafe { ptr::read(raw_ptr) }.core.state
+}
```
Oppen, line 72: This is unsound due to the fact neither `T` nor `H` have any trait bounds to make sure the types actually have a matching layout. The layout is also not guaranteed to match either way, because Rust specifies the compiler is free to reorder fields as it sees fit. I would at least check (if possible, at compile-time) that the `size_of` matches, for lack of a better canary.

Oppen: This can be somewhat improved by using [decompose](https://docs.rs/digest/0.10.6/digest/core_api/struct.CoreWrapper.html#method.decompose). Since we only care about getting `core.state` here, using `decompose` reduces the chances of mistakes by using the original `struct` instead for the first dereference.
It would look like this:
```rust
let core = hasher.decompose().0;
let raw_ptr = &core as *const _ as *const Sha3State<T>;
unsafe { ptr::read(raw_ptr) }.state
```

MarcosNicolau: Done!

PR #174, opened 2024-08-07
Address various issues

### LINKED ISSUE #113: Remove `unreachable` from `get_forward_memory_pointer`
On `get_forward_memory_pointer` under `far_call.rs` we are using the `unreachable` macro when the pointer source is of type `Forwarded`. We should first check that this branch is effectively unreachable, and then we should remove it in some way.

### LINKED ISSUE #114: Remove `panic` from `Predicate::from` method
Currently, we are panicking if input value does not correspond to any of the defined predicates. We should return an error instead and handle it in a proper way when this method is called.

MarcosNicolau:
This pr addresses various small issues: 

- Closes #114 
- Closes #113


Oppen (approved):

jorbush (approved):
LGTM

PR #177, opened 2024-08-08
Evm interpreter experimental

MarcosNicolau:
Experimental and very dirty solution for evm interpreter that fixes all tests under `solidity/ethereum/interpreter`. Follows #160. In order to be able to pass the test, you need to pull the changes from this [branch](https://github.com/lambdaclass/era-compiler-tester/pull/9) in the `era-compiler-tester`.

While this solution does address the test cases, the implementation itself is incorrect and should not be merged. To be more specific, contracts are being deployed directly within the `far_call` opcode, which is not correct. We need to implement a tracer or another appropriate method to handle this correctly. I'm opening this PR to provide visibility on how contracts may be deployed and to facilitate discussion.


MarcosNicolau:
Closing in favor of #160.

PR #187, opened 2024-08-12
Various gas fixes

MarcosNicolau:
Various fixes related to how we were managing the gas consumption:
- `inexplicit_panic` and `panic_from_far_call` count as instructions and we should be decreasing the gas base cost for running them as well.
- When pushing a new `far_call` frame if the bytecode was for the evm interpreter, we were returning back not only the left gas but also the stipend one.

### DIFF
--- a/src/call_frame.rs
+++ b/src/call_frame.rs
@@ -9,6 +9,7 @@ pub struct CallFrame {
     pub pc: u64,
     pub transient_storage_snapshot: SnapShot,
     pub gas_left: Saturating<u32>,
+    pub stipend: Saturating<u32>,
     pub exception_handler: u64,
     pub sp: u32,
     pub storage_snapshot: SnapShot,
@@ -43,7 +44,8 @@ impl Context {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         program_code: Vec<U256>,
-        gas_stipend: u32,
+        new_context_gas: u32,
+        stipend: u32,
         contract_address: Address,
         code_address: Address,
         caller: Address,
@@ -58,7 +60,8 @@ impl Context {
     ) -> Self {
         Self {
             frame: CallFrame::new_far_call_frame(
-                gas_stipend,
+                new_context_gas,
+                stipend,
                 exception_handler,
                 storage_snapshot,
                 transient_storage_snapshot,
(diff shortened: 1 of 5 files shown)


jorbush (approved):
LGTM

JulianVentura:
I have some questions regarding these fixes. Are we fixing some tests with these changes? If not, then how are we sure they are really necessary?

MarcosNicolau:
The reasons seem quite clear. When comparing the `gas_left` field during the execution of the interpreter tests, it was clear that the `ret` operation was adding a lot more more gas to the new frame due to the stipend. Even though this issue might not cause any test failures right now, given that the issue is pretty evident, I would not wait until for a potential future test failure, which might never occur.

gianbelinche:
A good way of testing this would be using the `benchmark-analyzer` of the compiler tester with the interpreter tests
The idea would be the following:
Run:
`cargo run --verbose --features lambda_vm --release --bin compiler-tester -- --path tests/solidity/complex/interpreter/test.json --mode 'Y+M3B3 0.8.25' --benchmark lambda_vm.json`
Then run:
`cargo run --verbose --release --bin compiler-tester -- --path tests/solidity/complex/interpreter/test.json --mode 'Y+M3B3 0.8.25' --benchmark vm1.json`

This will generate 2 json files with data for each interpreter test, we care about ergs, running:
`cargo run --release --bin benchmark-analyzer -- --reference vm1.json --candidate lambda_vm.json`
Will output a table like this

╔═╡ Size (-%) ╞═════╡ EVMInterpreter M3B3 ╞═╗
║               Mean                                                   0.000 ║
║               Best                                                     0.000 ║
║              Worst                                                   0.000 ║
║                Total                                                  0.000 ║
╠═╡ Cycles (-%) ╞═══╡ EVMInterpreter M3B3 ╞═╣
║                Mean                                                4.834 ║
║                Best                                                   7.022 ║
║               Worst                                                 -1.611 ║
║                Total                                                  5.573 ║
╠═╡ Ergs (-%) ╞═════╡ EVMInterpreter M3B3 ╞═╣
║               Mean                                                  -3.166 ║
║                Best                                                    0.000 ║
║               Worst                                                 -6.275 ║
║               Total                                                   -3.573 ║
╠═╡ Ergs/gas ╞══════╡ EVMInterpreter M3B3 ╞═╣
║              ADD                                                     67.667 ║
║               MUL                                                   39.400 ║
║              SUB                                                     67.667 ║
.
.
.
║            SWAP16                                            66.000 ║
║            RETURN                                                    inf ║
║             REVERT                                                   inf ║
╠═╡ Ergs/gas (-%) ╞═╡ EVMInterpreter M3B3 ╞═╣
║                 ADD                                              29.021 ║
║                 MUL                                               31.119 ║
║                  SUB                                              29.021 ║
.
.
.
║                 SWAP16                                       38.700 ║
╚════════════════════════════════╝

We care about `ergs/gas` which indicates the relation between the ergs spent by the eravm and the gas spent by the interpreter, since the gas should be the same in both, this will tell us if the ergs spent is correct, if all opcodes have `ergs/gas` equal 0 then we are calculating ergs correctly

Have in mind that we may have to change some things in compiler tester to let it know the ergs we spent

Oppen (approved):

PR #188, opened 2024-08-13
Docs `Bootloader`

jorbush:
Adds documentation of `Bootloader`, its relationship to EraVM, and explains Decommit`s relationship to it.


docs/general.md
```diff
@@ -534,4 +584,82 @@ add 8,r0,r1
...
+
+#### Bootloader Interaction with `decommit_code_hash`
+
+The `decommit_code_hash` function is essential for how the Bootloader handles contracts and their associated code within the EraVM system.
+
+1. **Contract Identification**: The Bootloader identifies the system's deployer contract address, generating a `storage_key` to access the contract's code in storage:
+    ```rust
+    let deployer_system_contract_address = Address::from_low_u64_be(DEPLOYER_SYSTEM_CONTRACT_ADDRESS_LOW as u64);
+    let storage_key = StorageKey::new(deployer_system_contract_address, address_into_u256(address));
+    ```
```
gianbelinche, line 625: These sections (until 5) are a bit odd, if I dind't know the bootloader I would think it is written in Rust, maybe specify a little more that that code is from the EraVM `decommit_code_hash` and how does the Bootloader access them

gianbelinche (changes requested):
Change the base branch to the one you based this PR from, when the precompile PR is merged it will change to main automatically, so we only see the changes specific to this PR

docs/general.md
jorbush: Just before it says it is from EraVM. Should I add anything else?
> The decommit_code_hash function is essential for how the Bootloader handles contracts and their associated code within the EraVM system.
When you said _how does the Bootloader access them_, do you mean the contracts or the hash of the code generated by this function?

gianbelinche: What I mean is since it says 
`The Bootloader identifies the system's deployer contract address, generating a storage_key to access the contract's code in storage:`
It seems like the bootloader is doing that in Rust, when its the function `decommit_code_hash` that is doing that
With _how does the Bootloader access them_ i meant how does it access the `decommit_code_hash` function

jorbush: you are right this is wrong

docs/general.md
```diff
@@ -534,4 +534,82 @@ add 8,r0,r1
...
+        _ if code_info == U256::zero() => try_default_aa.ok_or(StorageError::KeyNotPresent)?,
+        _ => return Err(EraVmError::IncorrectBytecodeFormat),
+    };
+    ```
+5. **Returning the Code Hash**: Finally, it prepares the code hash that the Bootloader will use for executing the contract:
+    ```rust
+    code_info_bytes[1] = 0;
+    Ok(U256::from_big_endian(&code_info_bytes))
+    ```
+This hash is critical, as it corresponds to the contract code executed during block construction.
```
IAvecilla, line 615: I would also add a section explaining how the hooks work, as it seems that there’s no documentation available on this topic. There is a brief explanation about the [hook pointer](https://docs.zksync.io/zk-stack/components/zksync-evm/bootloader#vm-hook-pointers) in the bootloader that might be useful.

To briefly explain how it works: basically, the bootloader uses a function call, [`setHook`](https://github.com/matter-labs/era-contracts/blob/8670004d6daa7e8c299087d62f1451a3dec4f899/system-contracts/bootloader/bootloader.yul#L3932), that writes to a specific part of the memory reserved to indicate that a hook is present. The VM executes this store to memory and [checks if it has been written in that specific part](https://github.com/matter-labs/vm2/blob/9a38900d7af9b1d72b47ce3be980e77c1239a61d/src/instruction_handlers/heap_access.rs#L117). If it has, the VM stores the hook type (there are several types) and the instruction to execute after the hook is handled. The VM will indicate that execution has stopped, and this will be [handled on the zksync-era side](https://github.com/matter-labs/zksync-era/blob/10674620d1a04333507ca17b9a34ab3cb58846cf/core/lib/multivm/src/versions/vm_fast/vm.rs#L111). Once the hook is handled, execution resumes, with the VM using the previously saved instruction pointer.

I’m not sure if it’s necessary to detail all the different parts involved in the hook processing, but that’s a general overview.

IAvecilla (changes requested):

docs/general.md
```diff
@@ -534,4 +534,82 @@ add 8,r0,r1
...
+
+It validates and executes transactions, handles errors effectively, and integrates with EraVM for seamless block construction.
+
+### Core Functions
+1. **Transaction Lifecycle Management**:
+    - **Validation**: Validates each transaction from the server to meet the required criteria.
+    - **Fee Processing**: Calculates and deducts the necessary transaction fees.
+    - **Execution**: Executes transactions, forming the new block.
+
+    Simplified workflow:
```
fkrause98, line 547: Can we state a bit more that this is a high-level idea on how the boot loader works? For clarity's sake.

fkrause98 (approved):

docs/general.md
```diff
@@ -534,4 +534,82 @@ add 8,r0,r1
 
 ## Bootloader
 
-Operator execution (transactions come in, get executed on the bootloader, state is suspended until new transaction shows up).
+The **Bootloader** is a system contract that orchestrates the construction and validation of new blocks, acting as a bridge between the EraVM and the external server to ensure accurate transaction processing.
+
+It validates and executes transactions, handles errors effectively, and integrates with EraVM for seamless block construction.
```
Oppen, line 539: I think we should mention why we need this to be a contract rather than part of the server functionality in the first place: to prove the execution of a batch we can't just prove the execution of the individual transactions, but rather we need to prove their sequencing, scheduling, loading and fees. This means we need to run all of this on top of the VM to make it provable, and the easiest way to do that is by writing it as a contract.

jorbush: I have added this information. Thank you!

docs/general.md
jorbush: Fixed! I added a clarifying line just before this.

docs/general.md
jorbush: Thank you! I have added the section, very helpful your resources and explanation.

PR #190, opened 2024-08-13
fix: code-page reads must always succeed

### LINKED ISSUE #189: Panic on code_page read
The spec states:
> Code pages
> A code page contains 216216 instructions. They are not writable.
> On genesis, code pages are filled as follows:
>     The contract code is places starting from the address 0.
>     The rest is filled with a value guaranteed to decode as [invalid](https://matter-labs.github.io/eravm-spec/spec.html#invalid) instruction.

However, [our code](https://github.com/lambdaclass/era_vm/blob/main/src/address_operands.rs#L84) only allocates space for the actual bytecode and accesses it via the `[]` operand, causing the VM to `panic` when reading past the end of the program.
We should have a `get` method or similar which returns `invalid` on that situation.
Triggered by the `zksync-era` benchmarks.

Oppen:
From the spec:
> Code pages
> A code page contains 2^16 instructions. They are not writable.
> On genesis, code pages are filled as follows:
> The contract code is places starting from the address 0.
> The rest is filled with a value guaranteed to decode as invalid
> instruction.

Our code before this fix panics when running the benchmarks due to this
discrepancy.

The applied fix creates a newtype for code pages that wraps the
`Vec<U256>` and returns `U256::zero()` when reading out of bounds, which
decodes to an invalid instruction, matching the spec and fixing the
`panic`.

EDIT: the test encoding exceeds the 2^16 range, so I removed that restriction.

Fixes #189

### DIFF
--- a/src/address_operands.rs
+++ b/src/address_operands.rs
@@ -81,7 +81,7 @@ pub fn address_operands_read(
                 ImmMemHandlerFlags::UseCodePage => {
                     let (src0, src1) = reg_and_imm_read(vm, opcode);
 
-                    let res = vm.current_context()?.code_page.get(src0.value.low_u16());
+                    let res = vm.current_context()?.code_page.get(src0.value.low_u16() as usize);
                     (TaggedValue::new_raw_integer(res), src1)
                 }
             }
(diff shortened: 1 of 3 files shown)


fkrause98 (approved):

fkrause98:
This PR also fixes the benches from zskync-era, so please also test this:
1. Go to our fork of zksync-era and checkout to `era_vm_integration_v2`.
2. Run: `cd etc/contracts-test-data/ && yarn && yarn build`

MarcosNicolau (approved):

PR #193, opened 2024-08-14
Add decommited_hashes to ContractStorage

MarcosNicolau:
Adds decommited hashes to `ContractStorage` trait.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "era-compiler-tester"]
 	path = era-compiler-tester
 	url = https://github.com/lambdaclass/era-compiler-tester.git
-	branch = zksync-era-tests
+	branch = decommited-hashes
--- a/src/op_handlers/far_call.rs
+++ b/src/op_handlers/far_call.rs
@@ -235,7 +235,7 @@ pub fn far_call(
 
     // Unlike all other gas costs, this one is not paid if low on gas.
     if decommit_cost < vm.gas_left()? {
-        vm.decrease_gas(decommit_cost);
+        vm.decrease_gas(decommit_cost)?;
     }
 
     let program_code = contract_storage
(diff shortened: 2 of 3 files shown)


.gitmodules
```diff
@@ -1,4 +1,4 @@
 [submodule "era-compiler-tester"]
 	path = era-compiler-tester
 	url = https://github.com/lambdaclass/era-compiler-tester.git
-	branch = zksync-era-tests
+	branch = decommited-hashes
```
juan518munoz, line 4: shouldn't we keep the submodule branch as is? It would be helpful for others if we link the corresponding PR in `era-compiler-tester` in the description, so they know they should be seen in tandem.

src/op_handlers/far_call.rs
```diff
@@ -193,10 +194,15 @@ fn decommit_code_hash(
         // Invalid
         _ => return Err(EraVmError::IncorrectBytecodeFormat),
     };
+    let code_key = U256::from_big_endian(&code_info_bytes);
+    let cost = if contract_storage.decommited_hashes()?.contains(&code_key) {
+        0
+    } else {
+        let code_length_in_words = u16::from_be_bytes([code_info_bytes[2], code_info_bytes[3]]);
```
Oppen, line 201: I think this will break for test encoding.

src/store.rs
```diff
@@ -39,21 +40,23 @@ impl InitialStorage for InitialStorageMemory {
 }
 
 pub trait ContractStorage: Debug {
-    fn decommit(&self, hash: U256) -> Result<Option<Vec<U256>>, StorageError>;
-    fn hash_map(&self) -> Result<HashMap<U256, Vec<U256>>, StorageError>;
+    fn decommit(&mut self, hash: U256) -> Result<Option<Vec<U256>>, StorageError>;
+    fn decommited_hashes(&self) -> Result<HashSet<U256>, StorageError>;
```
Oppen, line 44: I'd suggest the following methods instead:
```suggestion
    fn decommit(&'a mut self, hash: U256) -> Result<Option<&'a [U256]>, StorageError>;
    fn contains_decommited_hash(&self, hash: U256) -> Result<bool, StorageError>;
```

MarcosNicolau:
Closing this one in favor of #202.

Issue #196, opened 2024-08-14
Create struct to encapsulate all rollbackable fields


MarcosNicolau:
I’ve been working on implementing pubdata and refunds(see [here](https://github.com/lambdaclass/era_vm/pull/195)) to fix several integration tests for zksync-era with our VM. This requires adding several new fields (I believe I am even missing some):
```rust
pubdata: i32
pubdata_costs: Vec<i32>
refunds: Vec<i32>
read_storage_keys: Vec<StorageKey>
write_storage_keys: Vec<StorageKey>
```

All these fields need to be rollbackable. So far, only `state_storage` and `transient_storage` have been rollbacked, and so we were managing them by passing a snapshot to new frames, and when reverting we would just set the `state_storage` and `transient_storage` to their respective snapshot. 

The problem is, with all this new fields, we’d be passing a significant number of parameters to each opcode and frame. Additionally, whenever we need to revert a change, we’d have to update all relevant places to include each new field.

Before adding new features, I think it’s best to refactor the code by creating a struct that encapsulates all these fields and includes a rollback method that reverts each field, and a snapshot method that creates a copy of each field. This struct would be passed to each frame via the snapshot method, and in the case of any `ret.revert` or `ret.panic`, a rollback would be performed. This way, the opcodes would receive:

```rust
exec_opcode(&mut vm.state, &opcode, &mut rollbackable_struct)
```

This will also require an update to the integration on zksync-era, but it’s better to do it now while there are only a few fields rather than waiting until there are many more.

Additionally, there are other fields we should consider making rollbackable:
- events
- l2_l1_logs

MarcosNicolau:
Btw, they are doing something similar but with a `WorldDiff` struct(see [here](https://github.com/matter-labs/vm2/blob/master/src/world_diff.rs)). What I am proposing is much more simpler, without all the methods they have(most of them are only used once across the whole codebase), only a struct with similar fields but whose only responsibility would be to rollback each field and create snapshots.

juan518munoz:
I believe that an initial implementation akin to `WorldDiff` would be a good idea.

- `state_storage` and `transient_storage` must go inside of it.
- Implement the struct (and it's fields) with at least one test that makes use of it, we shouldn't push a struct with unused fields just in case, let's keep it as simple as possible.

One question:
```rs
exec_opcode(&mut vm.state, &opcode, &mut rollbackable_struct)
```
[...]

PR #197, opened 2024-08-14
Refactor rollbacks

MarcosNicolau:
This refactor abstract rollbacks to support an easier integration with new rollbackable fields. For that a new `World` struct is created to encapsulate all the those fields. See more #196. Link to the respective integrations prs:
- [era-compiler-tester](https://github.com/lambdaclass/era-compiler-tester/pull/17)
- [zksync-era](https://github.com/lambdaclass/zksync-era/pull/232)


juan518munoz (approved):
LGTM

src/op_handlers/far_call.rs
```diff
@@ -240,7 +239,9 @@ pub fn far_call(
         .checked_add(stipend)
         .expect("stipend must not cause overflow");
 
-    let program_code = contract_storage
+    let program_code = world
+        .contracts_storage
+        .borrow_mut()
```
fkrause98, line 244: Please put this into a method, I know it can be a bit more bureaucratic, but this can panic if for some reason the reference holds from another place in the codebase, putting this code inside a method guarantees the reference is dropped. Furthermore, we only want to use the storage for decommit, so this could be simply a 'decommit' method and be done with it.

fkrause98: You will also probably have to clone the vec with the program code, I think the cost is worth the hassle since the alternative is panicking.

src/op_handlers/log.rs
```diff
@@ -3,78 +3,75 @@ use u256::U256;
...
     let value = vm_state.get_register(opcode.src1_index).value;
     let is_service = opcode.imm0 == 1;
-    state_storage.record_l2_to_l1_log(L2ToL1Log {
+    world.record_l2_to_l1_log(L2ToL1Log {
         key,
         value,
         is_service,
         address: vm_state.current_context()?.contract_address,
         shard_id: 0,
         tx_number: vm_state.tx_number as u16,
```
fkrause98, line 74: Not from this PR, but, why tx_number is u16 but the VM struct is u64? This can panic, how are we checking this cast is safe?

src/op_handlers/opcode_decommit.rs
```diff
@@ -20,7 +20,9 @@ pub fn opcode_decommit(
 
     vm.decrease_gas(extra_cost)?;
 
-    let code = storage
+    let code = world
+        .contracts_storage
+        .borrow()
```
fkrause98, line 25: Same comment as above, this borrow will panic if there is an existing mutable reference to code, or a new one is created, let's wrap this in the 'decommit' method I proposed above, or something else, let's just make sure we drop the ref.

src/world.rs
```diff
@@ -0,0 +1,124 @@
...
+
+    fn rollback(&mut self, snapshot: Self::Snapshot) {
+        self.storage_changes.rollback(snapshot.storage_changes);
+        self.transient_storage.rollback(snapshot.transient_storage);
+        self.l2_to_l1_logs.rollback(snapshot.l2_to_l1_logs);
+        self.events.rollback(snapshot.events)
+    }
+
+    fn snapshot(&self) -> Self::Snapshot {
+        Self::Snapshot {
```
fkrause98, line 117: This can be expensive since we're cloning everything every time we do a far_call, have you considered wrapping this in a CoW pointer?

cc @Oppen, I'd like to hear your take on this.

fkrause98 (changes requested):
Looks good, and great job, but I want do discuss 2 things mostly:
- RefCell borrows should be done inside methods so we're 100% certain the ref is dropped, 
   and we don't panic.
- Snapshots cloning, this may be not that critical, but I'd like to hear @Oppen's input on this.

src/op_handlers/log.rs
MarcosNicolau: Mmm... I don't know about that, maybe it is worth opening an issue.

src/world.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    pub fn record_l2_to_l1_log(&mut self, msg: L2ToL1Log) {
+        self.l2_to_l1_logs.entries.push(msg);
+    }
+
+    pub fn record_event(&mut self, event: Event) {
+        self.events.entries.push(event);
+    }
+
+    pub fn decommit(&mut self, hash: U256) -> Result<Option<Vec<U256>>, StorageError> {
```
MarcosNicolau, line 95: Btw, this is marked as mutable on purpose, because when adding pubdata, refunds and decommited hashes we'll need to change some fields.

src/lib.rs
```diff
@@ -16,4 +16,6 @@ pub mod vm;
 pub use opcode::Opcode;
 pub use state::VMState;
 pub use vm::EraVM;
+mod rollbacks;
+pub mod world;
```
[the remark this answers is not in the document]
MarcosNicolau: Yes we do use it on the compiler-tester and zksync-era, for the `Event` and the `L2ToL1Log` structs.

src/op_handlers/log.rs
fkrause98: [Sure](https://github.com/lambdaclass/era_vm/issues/198)

fkrause98 (approved):

src/world.rs
```diff
@@ -0,0 +1,128 @@
...
+pub struct Event {
+    pub key: U256,
+    pub value: U256,
+    pub is_first: bool,
+    pub shard_id: u8,
+    pub tx_number: u16,
+}
+
+#[derive(Debug)]
+pub struct World {
```
IAvecilla, line 28: I wouldn't call this `World` as it might lead to confusion since this struct includes not only aspects related to the general state of the chain but also some internal VM state. Could we change the name to smth more related to storage? Perhaps `StateStorage` or `VMStorage` would be more appropriate. Additionally, we should update the name of `WorldSnapshot` to be consistent with this change. Feel free to propose other names if you have a different idea—these were just the first that came to mind.

PR #206, opened 2024-08-19
Optimization: Modify `Heap` read and write related functions (wip)

JulianVentura:
## Description

This PR improves vm performance by updating `Heap` read and write related functions.

### Inspiration

We are currently iterating and applying `U256` operations with a `for` loop on some `Heap` functions, like `read`. It works, but it's not very efficient. We may replace this with functions like `U256::from_big_endian` instead.

### Changes

* `Heap::store`
* `Heap::read`
* `Heap::expanded_read`
* `Heap::read_unaligned_from_pointer`

### Results

* Up to 45% improvement on bench `heap_read_write`.
* Around 5% improvement on the other benches.
* Around 11% improvement on execution time running all but `event_spam` benches.


JulianVentura:
Close because moved to #216

PR #208, opened 2024-08-20
Zksync era integration tests

MarcosNicolau:
This pr introduces several changes and fixes to the vm:
- Fixes `Contex(IncrementTxNumber)` opcode: it was not clearing the transient storage.
- Fixes the `decommit` opcode: it was not accounting if the contract had already been decommitted.
- Various fixes to the `farcall` opcode: adds mandated gas, handles decommit cost err, and fixes the params of mimic call
- Fix gas return on reverts and panics: on `far_calls` we were not supposed to add the `evm_interpreter` stipend gas to the new frame.
- Implements external rollbacks: this type of rollbacks are only called from the bootloader and they create a snapshot of the whole vm(execution and whole state).

All of this changes have been made to finish the integration with the bootloader. See [here](https://github.com/lambdaclass/zksync-era/pull/234).

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -29,6 +29,7 @@ pub struct Heap {
 }
 
 #[derive(Debug, Clone, PartialEq)]
+// represents the vm execution state
 pub struct Execution {
     // The first register, r0, is actually always zero and not really used.
     // Writing to it does nothing.
@@ -51,6 +52,7 @@ pub struct Execution {
 }
 
 #[derive(Debug, Clone, PartialEq)]
+// a saved state of the vm execution
 pub struct ExecutionSnapshot {
     pub registers: [TaggedValue; 15],
     pub flag_lt_of: bool,
@@ -318,7 +320,7 @@ impl Execution {
     }
 
     pub fn increase_gas(&mut self, to_add: u32) -> Result<(), EraVmError> {
-        self.current_frame_mut()?.gas_left += to_add;
+        self.current_frame_mut()?.gas_left += Saturating(to_add);
         Ok(())
     }
 
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -50,6 +50,22 @@ pub struct Execution {
     pub use_hooks: bool,
 }
 
+#[derive(Debug, Clone, PartialEq)]
+pub struct ExecutionSnapshot {
+    pub registers: [TaggedValue; 15],
+    pub flag_lt_of: bool,
+    pub flag_gt: bool,
+    pub flag_eq: bool,
+    pub running_contexts: Vec<Context>,
+    pub tx_number: u64,
+    pub heaps: Heaps,
+    pub register_context_u128: u128,
+    pub default_aa_code_hash: [u8; 32],
+    pub evm_interpreter_code_hash: [u8; 32],
+    pub hook_address: u32,
+    pub use_hooks: bool,
+}
+
 impl Execution {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
@@ -321,6 +337,37 @@ impl Execution {
 
     pub fn in_far_call(&self) -> bool {
         self.running_contexts.len() > 1
+    }
+
+    pub fn snapshot(&self) -> ExecutionSnapshot {
+        ExecutionSnapshot {
+            default_aa_code_hash: self.default_aa_code_hash,
+            evm_interpreter_code_hash: self.evm_interpreter_code_hash,
+            flag_eq: self.flag_eq,
+            flag_gt: self.flag_gt,
+            flag_lt_of: self.flag_lt_of,
+            heaps: self.heaps.clone(),
+            hook_address: self.hook_address,
+            register_context_u128: self.register_context_u128,
+            registers: self.registers.clone(),
+            running_contexts: self.running_contexts.clone(),
+            tx_number: self.tx_number,
+            use_hooks: self.use_hooks,
+        }
+    }
+    pub fn rollback(&mut self, snapshot: ExecutionSnapshot) {
+        self.default_aa_code_hash = snapshot.default_aa_code_hash;
+        self.evm_interpreter_code_hash = snapshot.evm_interpreter_code_hash;
+        self.flag_eq = snapshot.flag_eq;
+        self.flag_gt = snapshot.flag_gt;
+        self.flag_lt_of = snapshot.flag_lt_of;
+        self.heaps = snapshot.heaps;
+        self.hook_address = snapshot.hook_address;
+        self.register_context_u128 = snapshot.register_context_u128;
+        self.registers = snapshot.registers;
+        self.running_contexts = snapshot.running_contexts;
+        self.tx_number = snapshot.tx_number;
+        self.use_hooks = snapshot.use_hooks;
     }
 }
 
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -18,16 +18,17 @@ pub const CALLDATA_HEAP: u32 = 1;
 pub const FIRST_HEAP: u32 = 2;
 pub const FIRST_AUX_HEAP: u32 = 3;
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Stack {
     pub stack: Vec<TaggedValue>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 pub struct Heap {
     heap: Vec<u8>,
 }
-#[derive(Debug, Clone)]
+
+#[derive(Debug, Clone, PartialEq)]
 pub struct Execution {
     // The first register, r0, is actually always zero and not really used.
     // Writing to it does nothing.
@@ -49,8 +50,6 @@ pub struct Execution {
     pub use_hooks: bool,
 }
 
-// Totally arbitrary, probably we will have to change it later.
-pub const DEFAULT_INITIAL_GAS: u32 = 1 << 16;
 impl Execution {
     #[allow(clippy::too_many_arguments)]
     pub fn new(
@@ -63,6 +62,7 @@ impl Execution {
         evm_interpreter_code_hash: [u8; 32],
         hook_address: u32,
         use_hooks: bool,
+        initial_gas: u32,
     ) -> Self {
         let mut registers = [TaggedValue::default(); 15];
         let calldata_ptr = FatPointer {
@@ -76,7 +76,7 @@ impl Execution {
 
         let context = Context::new(
             program_code.clone(),
-            u32::MAX - 0x80000000,
+            initial_gas,
             contract_address,
             contract_address,
             caller,
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -87,6 +87,7 @@ impl Execution {
             context_u128,
             StateSnapshot::default(),
             false,
+            0,
         );
 
         let heaps = Heaps::new(calldata);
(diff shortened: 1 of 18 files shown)


src/op_handlers/far_call.rs
```diff
@@ -240,16 +246,26 @@ pub fn far_call(
...
     } = far_call_params_from_register(src0, vm)?;
 
+    let mandated_gas = if abi.is_system_call && src1.value == ADDRESS_MSG_VALUE.into() {
+        MSG_VALUE_SIMULATOR_ADDITIVE_COST
+    } else {
+        0
+    };
+
+    // mandated gas can surprass the 64/63 limit
+    let ergs_passed = ergs_passed + mandated_gas;
```
fkrause98, line 256: I don't understand, how are we ensuring we don't surpass the 64/63 limit?

src/op_handlers/ret.rs
```diff
@@ -63,7 +63,7 @@ pub fn ret(
         vm.clear_registers();
         vm.set_register(1, result);
         let previous_frame = vm.pop_frame()?;
-        vm.current_frame_mut()?.gas_left += previous_frame.gas_left;
+        vm.current_frame_mut()?.gas_left += previous_frame.gas_left.0 - previous_frame.stipend;
```
fkrause98, line 66: Let's use the decrease_gas method here.
There's nothing wrong with exposing the zero field, but we must do increase and decrease of gas with their corresponding methods to avoid over/underflows.

src/rollbacks.rs
```diff
@@ -51,16 +64,24 @@ impl<T: Clone> RollbackableVec<T> {
...
+        &self,
+        snapshot: <RollbackableVec<T> as Rollbackable>::Snapshot,
+    ) -> &[T] {
+        &self.entries[snapshot..]
+    }
 }
 
 impl<T: Clone> Rollbackable for RollbackableVec<T> {
-    type Snapshot = Vec<T>;
+    type Snapshot = usize;
```
fkrause98, line 77: Let's add some docs/comments on why this is a size, I understand that this is an index to the snapshot, but let's do it for clarity's sake.

src/state.rs
```diff
@@ -74,9 +77,25 @@ impl VMState {
             read_storage_slots: RollbackableHashSet::<StorageKey>::default(),
             written_storage_slots: RollbackableHashSet::<StorageKey>::default(),
             decommitted_hashes: RollbackableHashSet::<U256>::default(),
+            initial_values: HashMap::default(),
         }
     }
 
+    pub fn reset(&mut self) {
```
fkrause98, line 84: If we're setting everything as default, we can simply do Default::default() here.

src/op_handlers/ret.rs
```diff
@@ -103,12 +104,11 @@ pub fn inexplicit_panic(vm: &mut Execution, state: &mut VMState) -> Result<bool,
...
```
fkrause98, line 107: Same as the previous comment

fkrause98 (changes requested):

src/op_handlers/far_call.rs
Oppen: I think the comment is mentioning this is an exception to that limit. Whether that's correct or not, no idea.
BTW, the limit is 63/64, right?

src/op_handlers/ret.rs
Oppen: Maybe we could make it a private member instead to avoid mistakes.

src/op_handlers/far_call.rs
MarcosNicolau: Yes, the mandated gas is an exception to the rule, the rules is enforced before though(see [here](https://github.com/lambdaclass/era_vm/blob/b512b315aae5be18b639c80f710be7a7408cdcab/src/op_handlers/far_call.rs#L106)). 

PD: yes the comment is wrong is actually wrong, it is 63/64, I'll make sure to update it.

src/op_handlers/ret.rs
MarcosNicolau: It would be a good idea to make it private. Not for this pr though, I think there are a lot of places that would require a refactor.

src/execution.rs
```diff
@@ -319,6 +340,37 @@ impl Execution {
...
+            flag_gt: self.flag_gt,
+            flag_lt_of: self.flag_lt_of,
+            heaps: self.heaps.clone(),
+            hook_address: self.hook_address,
+            register_context_u128: self.register_context_u128,
+            registers: self.registers,
+            running_contexts: self.running_contexts.clone(),
+            tx_number: self.tx_number,
+            use_hooks: self.use_hooks,
+        }
```
Oppen, line 358: I think this is the same as:
```suggestion
        self.clone()
```

src/execution.rs
```diff
@@ -50,6 +51,23 @@ pub struct Execution {
     pub use_hooks: bool,
```
Oppen, line 51: If you extract the program, then the `ExecutionSnapshot` becomes just a clone of this. Maybe we could move the program out and turn `ExecutionSnapshot` in a simple wrapper?

juan518munoz (approved):
LGTM

fkrause98 (changes requested):
Let's discus some things.

src/op_handlers/ret.rs
```diff
@@ -103,12 +109,16 @@ pub fn inexplicit_panic(vm: &mut Execution, state: &mut VMState) -> Result<bool,
         vm.clear_registers();
         vm.set_register(1, result);
         let previous_frame = vm.pop_frame()?;
-        vm.current_frame_mut()?.gas_left += previous_frame.gas_left;
+        vm.increase_gas(
```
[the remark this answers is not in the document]
MarcosNicolau: Great, didn't know that!

src/rollbacks.rs
```diff
@@ -14,12 +14,25 @@ pub struct RollbackableHashMap<K: Clone + Hash, V: Clone> {
...
     pub fn new() -> Self {
         Self {
             map: HashMap::new(),
         }
     }
+
+    pub fn get_logs_after_snapshot(
+        &self,
+        snapshot: <RollbackableHashMap<K, V> as Rollbackable>::Snapshot,
+    ) -> HashMap<K, (Option<V>, V)> {
```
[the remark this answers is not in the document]
MarcosNicolau: It would be the tuple this function returns, where the first value would be the value in the snapshot and the second the value of the current map. The first one is marked as optional because the key could have been a recent write and so it won't be contained in the snapshot. 

This is not used in the era_vm itself but in the `zksync-era` integration. We could return a struct to make it more declarative, the problem is that it would require a new pr on the `zksync-era` repo. Though, we could annotate this for a future refactor.

src/state.rs
```diff
@@ -194,28 +252,67 @@ impl VMState {
         self.events.entries.push(event);
     }
 
-    pub fn decommit(&mut self, hash: U256) -> Option<Vec<U256>> {
-        self.decommitted_hashes.map.insert(hash);
-        self.storage.borrow_mut().decommit(hash)
+    pub fn decommit(&mut self, hash: U256) -> (Option<Vec<U256>>, bool) {
+        let was_decommitted = !self.decommitted_hashes.map.insert(hash);
```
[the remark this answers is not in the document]
MarcosNicolau: When decommiting, it is useful to know whether the hash has already been decommited or not. This is used to calculate refunds and decommit costs.

src/state.rs
```diff
@@ -194,28 +252,67 @@ impl VMState {
...
+            })
+            .collect()
+    }
+
+    /// returns the values that have actually changed from the snapshot storage or the initial if the former does not exist
+    /// also, a flag is also retured indicating if the value existed on the initial storage
+    pub fn get_storage_changes_from_snapshot(
+        &self,
+        snapshot: <RollbackableHashMap<StorageKey, U256> as Rollbackable>::Snapshot,
+    ) -> Vec<(StorageKey, Option<U256>, U256, bool)> {
```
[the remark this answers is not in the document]
MarcosNicolau: Yes, it would be better to use a struct here, but again that would require a new pr on zksyn-era 🙃.

src/state.rs
```diff
@@ -235,9 +334,31 @@ impl Rollbackable for VMState {
...
             pubdata: self.pubdata.snapshot(),
-            pubdata_costs: self.pubdata_costs.snapshot(),
             paid_changes: self.paid_changes.snapshot(),
+        }
+    }
+}
+
+impl VMState {
+    pub fn external_rollback(&mut self, snapshot: FullStateSnapshot) {
+        self.rollback(snapshot.internal_snapshot);
```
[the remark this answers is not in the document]
MarcosNicolau: Not really, the can't fail. I've written some docs in this [pr](#208). I'll make sure to add some comments tho.

src/state.rs
```diff
@@ -118,19 +145,46 @@ impl VMState {
...
+        value
+    }
+
+    fn storage_read_inner(&mut self, key: &StorageKey) -> Option<U256> {
         match self.storage_changes.map.get(key) {
-            None => self.storage.borrow_mut().storage_read(key),
+            None => {
+                let cached_value = *self.initial_values.get(key).unwrap_or(&None);
+                if cached_value.is_none() {
+                    let value = self.storage.borrow_mut().storage_read(key);
```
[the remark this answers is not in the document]
MarcosNicolau: Yes, because we are adding a caching layer for the storage, to not query the db in the `zksync-era`, since we read a lot, specially when getting diffs.

src/state.rs
```diff
@@ -194,28 +252,67 @@ impl VMState {
...
+    pub storage_changes: <RollbackableHashMap<StorageKey, U256> as Rollbackable>::Snapshot,
+    pub transient_storage: <RollbackableHashMap<StorageKey, U256> as Rollbackable>::Snapshot,
+    pub l2_to_l1_logs: <RollbackableVec<L2ToL1Log> as Rollbackable>::Snapshot,
+    pub events: <RollbackableVec<Event> as Rollbackable>::Snapshot,
+    pub pubdata: <RollbackablePrimitive<i32> as Rollbackable>::Snapshot,
+    pub paid_changes: <RollbackableHashMap<StorageKey, u32> as Rollbackable>::Snapshot,
+}
+
+pub struct FullStateSnapshot {
+    pub internal_snapshot: StateSnapshot,
```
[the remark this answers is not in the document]
MarcosNicolau: I agree that names can be a bit confusing, I've added some new comments, and renamed `FullStateSnapshot` to `ExternalStateSnapshot`. Tell me if you think that is clearer 😁.

src/state.rs
```diff
@@ -118,19 +145,46 @@ impl VMState {
...
             value => value.copied(),
         }
     }
 
     pub fn storage_write(&mut self, key: StorageKey, value: U256) -> u32 {
         self.storage_changes.map.insert(key, value);
+
+        self.initial_values
+            .entry(key)
+            .or_insert_with(|| self.storage.borrow_mut().storage_read(&key)); // caching the value
```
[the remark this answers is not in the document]
MarcosNicolau: I think it does release the borrow, but I'll change it to be sure.

MarcosNicolau:
I’ve removed the cache for the initial storage to reduce confusion. We can consider re-adding it later if necessary.

src/state.rs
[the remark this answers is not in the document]
fkrause98: Let's add a comment clarifying this then.

fkrause98 (approved):

PR #212, opened 2024-08-22
Implement tracers

MarcosNicolau:
This pr adds a full implementation of tracers.


src/tracers/blob_saver_tracer.rs
```diff
@@ -106,7 +110,23 @@ impl Tracer for BlobSaverTracer {
...
 
         self.blobs.insert(key, as_words);
+    }
+
+    fn after_execution(
+        &mut self,
+        _opcode: &Opcode,
+        _execution: &mut Execution,
+        _state: &mut VMState,
+    ) {
```
juan518munoz, line 120: nit: add `todo!()` inside of functions that should have logic inside.

src/tracers/no_tracer.rs
```diff
@@ -0,0 +1,34 @@
+use super::tracer::Tracer;
+use crate::state::VMState;
+use crate::{execution::Execution, Opcode};
+
+#[derive(Default)]
+pub struct NoTracer {}
+
+impl Tracer for NoTracer {
```
juan518munoz, line 8: It's ok to leave functions empty inside of this `Tracer`.

juan518munoz (changes requested):
Looks ok, as mentiones before, we should add `todo!()` so we know later when we are trying to make use of a certain tracer, yet it is not implemented yet.

src/tracers/tracer.rs
```diff
@@ -1,11 +1,8 @@
...
-        &mut self,
-        _opcode: &Opcode,
-        _vm: &mut Execution,
-    ) -> Result<(), EraVmError> {
-        Ok(())
-    }
+    fn before_decoding(&mut self, execution: &mut Execution, state: &mut VMState);
+    fn after_decoding(&mut self, opcode: &Opcode, execution: &mut Execution, state: &mut VMState);
+    fn before_execution(&mut self, opcode: &Opcode, execution: &mut Execution, state: &mut VMState);
+    fn after_execution(&mut self, opcode: &Opcode, execution: &mut Execution, state: &mut VMState);
```
juan518munoz, line 7: Why are we dropping `Result` here?

juan518munoz (changes requested):

src/tracers/tracer.rs
MarcosNicolau: Because tracers are functions to be accepted from the outside word. The vm executes them but doesn't care about what they do, thus they shouldn't stop the execution or make the vm panic because, what would be the reason?

src/tracers/blob_saver_tracer.rs
MarcosNicolau: I've reimplemented the trait and now they have a default behavior of doing nothing, so we don't have to add them for every single tracer.

src/tracers/no_tracer.rs
MarcosNicolau: This tracer is meant to be used when the client doesn't provide a custom tracer.

src/tracers/tracer.rs
juan518munoz: > The vm executes them but doesn't care about what they do

This seems correct.

>  Thus they shouldn't stop the execution or make the vm panic

This too, but note that code like [this](https://github.com/lambdaclass/era_vm/pull/212/files#diff-15d730b3f5d168f25f65469c1db5d90300d85cd8c479aad0988850978dd8a772R28):
```rs
let (src0, src1) = address_operands_read(vm, opcode).unwrap();
```

Will stop the execution of out vm due to a panic.

MarcosNicolau: Oh yes, good catch, I'll make sure to change that 😁

src/tracers/blob_saver_tracer.rs
```diff
@@ -53,8 +48,12 @@ fn hash_evm_bytecode(bytecode: &[u8]) -> H256 {
...
 
 impl Tracer for BlobSaverTracer {
-    fn before_execution(&mut self, _opcode: &Opcode, vm: &mut Execution) -> Result<(), EraVmError> {
-        let current_callstack = vm.current_context()?;
+    fn before_execution(&mut self, _opcode: &Opcode, vm: &mut Execution, _state: &mut VMState) {
+        let current_callstack = vm.current_context();
+        if current_callstack.is_err() {
+            return;
+        }
+        let current_callstack = current_callstack.unwrap();
```
Oppen, line 56:
```suggestion
        let Ok(current_callstack) = current_callstack. else {
            return;
        };
```

src/tracers/blob_saver_tracer.rs
```diff
@@ -63,38 +62,43 @@ impl Tracer for BlobSaverTracer {
...
+        let heap = vm.heaps.get(ptr.page);
+        if heap.is_none() {
+            return;
+        }
+
+        let data = heap.unwrap().read_unaligned_from_pointer(&ptr);
+        if data.is_err() {
+            return;
+        }
+        let data = data.unwrap();
```
Oppen, line 84:
```suggestion
        let Ok(heap) = vm.heaps.get(ptr.page) else {
            return;
        };

        let Ok(data) = heap.read_unaligned_from_pointer(&ptr) else {
            return;
        };
```

src/tracers/print_tracer.rs
```diff
@@ -25,21 +25,23 @@ impl Tracer for PrintTracer {
         .unwrap();
 
         if matches!(opcode_variant, ZKOpcode::UMA(UMAOpcode::HeapWrite)) {
-            let (src0, src1) = address_operands_read(vm, opcode)?;
+            let (src0, src1) = address_operands_read(vm, opcode).unwrap();
```
Oppen, line 28:
```suggestion
            let Ok((src0, src1)) = address_operands_read(vm, opcode) else {
                return;
            };
```

src/tracers/blob_saver_tracer.rs
fkrause98: Do not commit this suggestion, it won't compile (mind the dot after the current_callstack)

src/tracers/blob_saver_tracer.rs
fkrause98: +1 to using let else.

src/tracers/print_tracer.rs
fkrause98: +1

src/tracers/tracer.rs
```diff
@@ -1,11 +1,26 @@
-use crate::{eravm_error::EraVmError, execution::Execution, Opcode};
+use crate::{execution::Execution, state::VMState, Opcode};
 
 pub trait Tracer {
+    fn before_decoding(&mut self, _execution: &mut Execution, _state: &mut VMState) {}
```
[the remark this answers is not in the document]
MarcosNicolau: Yes, the idea is to leave it blank, as the implementation is optional. Regarding testing, everything is being tested in zksync-era, where we have tracers like circuits_tracer. You can check it out [here](https://github.com/lambdaclass/zksync-era/pull/243).

src/tracers/last_state_saver_tracer.rs
```diff
@@ -38,10 +38,14 @@ impl Default for LastStateSaverTracer {
...
 
 impl Tracer for LastStateSaverTracer {
-    fn before_execution(&mut self, opcode: &Opcode, vm: &mut Execution) -> Result<(), EraVmError> {
+    fn before_execution(
+        &mut self,
+        opcode: &Opcode,
+        execution: &mut Execution,
+        _state: &mut VMState,
+    ) {
         if opcode.variant == Variant::Ret(RetOpcode::Ok) {
```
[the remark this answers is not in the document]
MarcosNicolau: This tracer isn't actually used anywhere. I don't think we can take ownership, given that it is implementing the Tracer trait, which takes a reference. Also, I think that would imply moving the value from the `era_vm` struct.

src/vm.rs
```diff
@@ -82,26 +82,30 @@ impl EraVM {
     }
 
     /// Run a vm program with a given bytecode.
-    pub fn run_program_with_custom_bytecode(&mut self) -> (ExecutionOutput, BlobSaverTracer) {
-        self.run_opcodes()
+    pub fn run_program_with_custom_bytecode(
+        &mut self,
+        tracer: Option<&mut dyn Tracer>,
```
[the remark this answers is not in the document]
MarcosNicolau: I agree with keeping the existing function intact to avoid breaking changes. I've added new functions, `run_program_with_test_encode_and_tracer` & `run_program_with_custom_bytecode_and_tracer`, names might be long but I don't think there is an alternative.

Let me know if this works!

src/tracers/blob_saver_tracer.rs
Oppen: Which dot :trollface:

Oppen (approved):

fkrause98 (approved):

PR #213, opened 2024-08-26
Implement statistics

MarcosNicolau:
This pr adds vm statistics to be used in the zksync-era integration, that includes:
- Collecting precompiles round counts
- Collect storages reads and writes cycles.
- Collect vm number of instructions run (or, as they call it, monotonic counter)
- Collect decommit cycles

This implementation is supported by the [circuits tracer tests](https://github.com/lambdaclass/zksync-era/pull/243/commits/8abf5c725da6050fdcbfa6eabe0fc7a2856046a5) in the zksync-era integration.

### DIFF
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1,5 +1,5 @@
 use lazy_static::lazy_static;
-use zkevm_opcode_defs::Opcode as Variant;
+pub use zkevm_opcode_defs::Opcode as Variant;
 use zkevm_opcode_defs::OpcodeVariant;
 use zkevm_opcode_defs::Operand;
 use zkevm_opcode_defs::CONDITIONAL_BITS_SHIFT;
--- a/src/opcode.rs
+++ b/src/opcode.rs
@@ -1,3 +1,5 @@
+use crate::eravm_error::EraVmError;
+use crate::eravm_error::OpcodeError;
 use lazy_static::lazy_static;
 pub use zkevm_opcode_defs::Opcode as Variant;
 use zkevm_opcode_defs::OpcodeVariant;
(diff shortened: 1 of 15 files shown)


Oppen:
Great idea! Two comments:
1. I haven't checked, so maybe it's already there but not mentioned in the description, both stats for far and near call depths would be useful.
2. This should probably be gated by a feature flag, since it may add significant overhead.

src/precompiles/ecrecover.rs
```diff
@@ -18,11 +18,13 @@ pub use zkevm_opcode_defs::sha3::Keccak256;
 
 use super::*;
 
+const NUM_ROUNDS: usize = 1;
```
IAvecilla, line 21: We have this constant defined for two different precompiles, but it has the same value in both cases. We could move it to the `statistics.rs` file and rename it to something like `DEFAULT_PRECOMPILE_NUM_ROUNDS` to have it in one place.

src/opcode.rs
```diff
@@ -1,14 +1,14 @@
+use crate::eravm_error::EraVmError;
+use crate::eravm_error::OpcodeError;
 use lazy_static::lazy_static;
-use zkevm_opcode_defs::Opcode as Variant;
+pub use zkevm_opcode_defs::Opcode as Variant;
```
IAvecilla, line 4: Was this change necessary for the `zksync-era` integration?

src/state.rs
```diff
@@ -268,7 +277,7 @@ impl VMState {
             .map
             .iter()
             .filter_map(|(key, value)| {
-                let initial_value = self.storage_read_inner(key);
+                let initial_value = self.storage.borrow_mut().storage_read(key);
```
IAvecilla, line 280: This is outside the scope, so it's not necessary to change it here. However, I'll add a comment to link this to a future issue. The `storage_read()` operation should not require a mutable reference.

IAvecilla (approved):

src/opcode.rs
MarcosNicolau: Yes, as the libraries used diverge, this way we don't have to add a new crate.

src/state.rs
MarcosNicolau: Noted, I'll make sure to add a comment to ensure we correct this in a future refactor!

fkrause98 (approved):

fkrause98 (approved):

PR #217, opened 2024-08-28
Optimization: Small changes on many functions V3 (wip)

JulianVentura:
## Description

This PR improves vm performance by making small changes to many functions.

### Changes

TODO

### Results

TODO

### DIFF
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -475,6 +475,17 @@ impl Heap {
         }
     }
 
+    pub fn store_multiple(&mut self, start_address: u32, values: Vec<U256>) {
+        let mut start = start_address as usize;
+        let end = (start + values.len() * 32).min(self.heap.len());
+        let max_to_store = (end - start) / 32;
+
+        for value in &values[..max_to_store] {
+            value.to_big_endian(&mut self.heap[start..(start + 32)]);
+            start += 32;
+        }
+    }
+
     pub fn read(&self, address: u32) -> U256 {
         let mut result = U256::zero();
 
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -390,12 +390,13 @@ impl Stack {
     }
 
     pub fn fill_with_zeros(&mut self, value: usize) {
-        for _ in 0..value {
-            self.stack.push(TaggedValue {
+        self.stack.resize(
+            self.stack.len() + value,
+            TaggedValue {
                 value: U256::zero(),
                 is_pointer: false,
-            });
-        }
+            },
+        );
     }
 
     pub fn get_with_offset(&self, offset: u16, sp: u32) -> Result<TaggedValue, StackError> {
--- a/src/execution.rs
+++ b/src/execution.rs
@@ -300,11 +300,12 @@ impl Execution {
         let pc = self.current_frame()?.pc;
         let raw_opcode = current_context.code_page.get(pc as usize / 4);
 
+        let [a, b, c, d] = raw_opcode.0;
         let raw_op = match pc % 4 {
-            3 => (raw_opcode & u64::MAX.into()).as_u64(),
-            2 => ((raw_opcode >> 64) & u64::MAX.into()).as_u64(),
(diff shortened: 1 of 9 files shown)


src/op_handlers/far_call.rs
```diff
@@ -190,23 +191,21 @@ fn decommit_code_hash(
...
         }
         // EOA: There is no code, so we return the default
-        _ if code_info == U256::zero() => try_default_aa.ok_or(StorageError::KeyNotPresent)?,
+        _ if code_info.is_zero() => try_default_aa.ok_or(StorageError::KeyNotPresent)?,
         // Invalid
         _ => return Err(EraVmError::IncorrectBytecodeFormat),
     };
 
-    code_info_bytes[1] = 0;
+    // let code_key = U256::from_big_endian(&code_info_bytes);
```
JulianVentura, line 199: Delete this comment

src/op_handlers/opcode_decommit.rs
```diff
@@ -52,11 +52,7 @@ pub fn opcode_decommit(
 
     let heap = vm.heaps.get_mut(id).ok_or(HeapError::StoreOutOfBounds)?;
 
-    let mut address = 0;
-    for value in code.iter() {
-        heap.store(address, *value);
-        address += 32;
-    }
```
Oppen, line 59: Seeing it starts with address 0, maybe we can just assign `code` to the new heap instead. This way we avoid all the copying and just move the whole vector.

src/execution.rs
```diff
@@ -474,6 +475,17 @@ impl Heap {
         }
     }
 
+    pub fn store_multiple(&mut self, start_address: u32, values: Vec<U256>) {
```
Oppen, line 478:
```suggestion
    pub fn store_multiple(&mut self, start_address: u32, values: &[U256]) {
```
This avoids the need to clone vectors if the data needs to be reused later.

src/op_handlers/opcode_decommit.rs
JulianVentura: Good idea!

src/execution.rs
JulianVentura: Thats right, it would be more efficient if we had to reuse the vector. I'll make the change.
Anyways, I'm curious about this particular case. It shouldn't be any performance difference if we don't want to use the vector after calling this function, right? Using a reference vs moving the value to the callee should be the same.

Oppen: It should be about the same, yes. The memcpy/load used by the move of `Vec` is slightly larger than it is for the slice (3 vs 2 machine words), but I would be surprised if it becomes relevant except in very edge cases.

JulianVentura: Great to know, thanks!

src/op_handlers/opcode_decommit.rs
JulianVentura: I actually don't know if there is an easy way of doing that, since `code` is of type `Vec<U256>` but `heap` is of type `Vec<u8>`.
I'll keep the code as is, but if you happen to know how to do that transformation it could be a good optimization.

MarcosNicolau: I am not saying to make this change now, but something to consider for a future pr: we could easily retrieve the code as a `Vec<u8>` by adding a new function to the `Storage` trait that decommits and returns it in that format. The reason it's easy is that in the current `zksync-era` storage, bytecodes are originally stored as `Vec<u8>`, and we are actually converting them to Vec<U256> (see [here](https://github.com/lambdaclass/zksync-era/blob/era_vm_integration_v2/core/lib/multivm/src/versions/era_vm/vm.rs#L724-L747)).

MarcosNicolau (approved):

src/op_handlers/opcode_decommit.rs
JulianVentura: Yep, if we are able to modify that trait (we are not following any API), then thats totally an option.

PR #218, opened 2024-08-28
Move benchmark contracts to era_vm

JulianVentura:
This PR moves custom made benchmark contracts sources from zksync-era to era_vm, so its easier to modify them.
It also adds a new benchmark result processing script made in python, which allows us to easily aggregate all benchmark execution times.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,19 +47,19 @@ define build_zk_contracts
 	$(1)
 endef
 
-$(ZKSYNC_L1_CONTRACTS)::
+$(ZKSYNC_L1_CONTRACTS):
 	$(call build_zk_contracts, yarn l1 build)
 
-$(ZKSYNC_L2_CONTRACTS)::
+$(ZKSYNC_L2_CONTRACTS):
 	$(call build_zk_contracts, yarn l2 build)
 
-$(ZKSYNC_SYS_CONTRACTS)::
+$(ZKSYNC_SYS_CONTRACTS):
 	$(call build_zk_contracts, yarn sc build:system-contracts)
 
-$(ZKSYNC_BOOTLOADER_CONTRACT)::
+$(ZKSYNC_BOOTLOADER_CONTRACT):
 	$(call build_zk_contracts, yarn sc build:bootloader)
 
-$(ZKSYNC_BENCH_TEST_DATA)::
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -84,6 +84,9 @@ bench-setup: submodules build_bench_contracts $(ZKSYNC_BENCH_TEST_DATA) $(ZKSYNC
 bench:
 	cd $(ZKSYNC_ROOT) && cargo bench --bench criterion "$(lambda|fast_vm|legacy)/(fibonacci|send)"
 
+check-flamegraph:
+	cd $(ZKSYNC_ROOT) && CARGO_PROFILE_BENCH_DEBUG=2 cargo flamegraph --root --bench criterion -- --bench --profile-time 5 "$lambda/fibonacci_rec^"
```
Oppen, line 88: I think the `$lambda` is converted to `ambda` for some reason. What does `$l` do here? Is the `$` even necessary?

bench_results.py
```diff
@@ -0,0 +1,55 @@
...
+        cleaned_content = re.sub(r'\n(\s+)time:', ' time:', content)
+        for line in cleaned_content.split('\n'):
+            match = re.match(pattern, line)
+            if match:
+                machine, lower_unit, mid_time, mid_unit, upper_unit = match.groups()
+                # Convert mid_time to milliseconds
+                mid_time_ms = convert_to_ms(float(mid_time), mid_unit)
+                machines[machine] = machines.get(machine, 0) + mid_time_ms
+
+    return machines
```
Oppen, line 30: `criterion` is capable of emitting JSON IIRC. Maybe that could save you parsing.

Oppen (approved):

Oppen:
Just to make sure, we should check the generated assembly for the contracts to confirm they do what we think they do.

Makefile
JulianVentura: No, it is not necessary actually. I had been using it without any problems though, so I think it works ok.

bench_results.py
JulianVentura: I didn't know that, it might be useful to improve the script. Thanks!

JulianVentura:
LGTM

fkrause98 (approved):

PR #221, opened 2024-09-03
Zk server compatibility

gianbelinche:
This PR makes the neccessary changes on era vm to be compatible with zksync-era

Related to:
https://github.com/lambdaclass/zksync-era/pull/253
https://github.com/lambdaclass/era-compiler-tester/pull/24


src/execution.rs
```diff
@@ -485,13 +499,19 @@ impl Heap {
...
         self.heap[address as usize]
     }
 
     pub fn read_from_pointer(&self, pointer: &FatPointer) -> U256 {
         let mut result = U256::zero();
         for i in 0..32 {
             let addr = pointer.start + pointer.offset + (31 - i);
+            if addr >= self.heap.len() as u32 {
+                continue;
+            }
```
Oppen, line 514: This can be done out of the loop, like this:
```rust
let start = pointer.start + offset;
let end = self.heap.len().min(start + 32);
for i in start..end {
    result |= U256::from(self.heap[addr as usize]) << (1 * 8);
}
result
```
Or quite possibly as a slice copy, if the `U256` provides a byte representation.

